PR #8, opened 2022-06-28
fix abbreviations - not catching capitals at end

amyconroy:
fixes: 

- [x] ensures that the first and last chars are capitals
- [x] where a quote slips through with the short form, this is removed
- [x] length of the abbreviation MUST be three chars or greater

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -149,3 +149,4 @@ lambda/determine_replacements_abbreviations/workspace.code-workspace
 *.xml
 legislation_provisions_extraction/test_judgments
 legislation_provisions_extraction/output
+abbreviation_extraction/abbreviations_matcher_test.py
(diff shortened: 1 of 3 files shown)


amyconroy:
@danielhoadley @enemsic I've added in a fix to catch the issues that Editha spotted, from the test judgment I ran now it truly only catches those that are capitals beginning/and last char as well as over three characters. 

Another potential limitation we can add is that the number of words in a long form must be greater than or equal to the number of characters in the short form. 

Ex: HRA / Human Rights Act has three chars/three words. 

This will stop the random issues that seem to pop up from coming through 

Ex: AC / Amy (where it doesn't pick up the Conroy)

enemsic (approved):

PR #58, opened 2023-01-10
Ontology: change root URI and remove hashes

robinchesterman:
This PR:

- changes the namespace of ontology URIs from `http://caselaw.nationalarchives.gov.uk/` to `http://caselaw.nationalarchives.gov.uk/def/caselaw/`
- removes some hash-based URIs left behind in the UK court system ontology


danielhoadley (approved):
These modification reflect the changes discussed and identified by the development team.

PR #74, opened 2023-01-18
Added bucket policy for vcite_enriched

ak-mdrxtech:
Signed-off-by: Adam King <adam.king@mdrx.tech>


ak-mdrxtech:
This looks to fail on a prod plan as the bucket doesn't exist there yet.

PR #124, opened 2023-01-23
Fix casing in court ontology

danielhoadley:
Apply snake-case to courts, e.g. `AdministrativeCourt` to `Administrative_Court`


robinchesterman (approved):
All looks good.

PR #146, opened 2023-02-16
Try to fix tests

dragon-dxw:
Trying to get the offline tests to run, let alone pass -- I think this might be all dead code since some of the "this doesn't parse" issues date from June 2022.


tests/test_xml_parser.py
```diff
@@ -56,14 +56,12 @@ def testXMLStrings(self):
                 <content>\
                     <p style="margin-left:0.50in;margin-right:0.00in;text-indent:0.5in">The present case is different.  The Sellers had a single obligation to provide and maintain the capacity to deliver gas from the Reservoirs at 130% of the TRDQ, whatever that was at the time.  They had no discretion as to how much daily gas they had to deliver.  As already noted, and the distinction is a subtle one, the power to change an obligation is not a discretion as to how to perform it.  Accordingly, once the obligation is identified and the breach established, the only remaining question is what if any loss has flowed from the breach.  There is no uncertainty about definition of the loss, though its quantification may be complex, depending on the circumstances.   </p>\
                 </content>\
-                </paragraph>\
-                <paragraph>\
```
anthonyhashemi, line 60: this test was failing because `parse_file` was iterating though all the content elements but it didnt find the second content element when in a separate paragraph

jacksonj04 (approved):
Approve even though tests aren't solid, on the basis that testing something and getting failures is better than testing nothing.

anthonyhashemi:
> 

I feel like a lot of the tests are still beneficial - they just needed updating. There are still a lot of tests that test multiple scenarios which may be redundant or at least need splitting up but can be done in a separate PR

anthonyhashemi (approved):

PR #152, opened 2023-02-27
Start fixing tests that depend on postgres to depend on test db

anthonyhashemi:
Here we make the tests use testing.postgres package instead of an actual local postgres db. Also simplify tests that do not need a db to not use one.

Thought about a way to avoid needing a db for the unit tests but it seems like the remaining tests with db access make sense to have access to a db.

There will be one more PR to fix some `filter_matches` tests in `tests/abbreviation_extraction_tests/test_abbreviations.py` which have issues for short form abbreviations which include lower case "the" as a prefix, e.g. "the VAT Act". After that all tests will be passing. Then will make a final PR to make the tests run in CI.

### DIFF
--- a/tests/caselaw_extraction_tests/test_case_citations.py
+++ b/tests/caselaw_extraction_tests/test_case_citations.py
@@ -19,7 +19,6 @@ from database.db_connection import get_matched_rule
 from replacer.replacer import replacer_caselaw
 from utils.helper import load_patterns
 
-
 CORRECT_CITATIONS = [
     "random text goes here random text goes here **[2022] UKUT 177 (TCC)",
     "[2022] 1 Lloyd's Rep 123.",
--- a/tests/caselaw_extraction_tests/test_case_citations.py
+++ b/tests/caselaw_extraction_tests/test_case_citations.py
@@ -1,22 +1,23 @@
-import sys
-import unittest
-
-from spacy.lang.en import English
-
-sys.path.append("./")
-from caselaw_extraction.correction_strategies import apply_correction_strategy
-from database.db_connection import (
-    close_connection,
-    create_connection,
-    get_matched_rule,
-)
-from replacer.replacer import replacer_caselaw
-from utils.helper import load_patterns
-
 """
     Testing the matching of the citations based on the data found in the rules.
     These are independent unit tests.
 """
+
+import sys
+import unittest
+
+import pandas as pd
+import psycopg2
+import testing.postgresql
+from spacy.lang.en import English
+from sqlalchemy import create_engine
+
+from caselaw_extraction.correction_strategies import apply_correction_strategy
+from database.db_connection import get_matched_rule
+from replacer.replacer import replacer_caselaw
+from utils.helper import load_patterns
+
+sys.path.append("./")
 
 CORRECT_CITATIONS = [
     "random text goes here random text goes here **[2022] UKUT 177 (TCC)",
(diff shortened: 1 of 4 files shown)


tests/caselaw_extraction_tests/test_case_citations.py
```diff
@@ -292,7 +280,7 @@ def test_correct_forms(self):
             citation_type, citation_match, canonical_form
         )
         assert corrected_citation == citation_match
-        assert year == "No Year"
+        assert year == ""
```
anthonyhashemi, line 283: just before mdrx handed over to use there was a change to leave year as empty string rather than "No Year" when there is no year in the original citation.

tests/requirements.txt
```diff
@@ -2,6 +2,8 @@ spacy==3.2.2
 en-core-web-sm @ https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.2.0/en_core_web_sm-3.2.0-py3-none-any.whl
 pandas==1.4.1
 psycopg2==2.9.5
-bs4==4.11.2
+beautifulsoup4==4.11.2
 spaczz==0.5.4
 lxml==4.9.2
+testing.postgresql==1.3.0
```
anthonyhashemi, line 8: added these to deal with the db stuff we added in the tests that require them

tests/caselaw_extraction_tests/test_case_citations.py
```diff
@@ -82,28 +83,29 @@ def mock_get_rules_total(db_conn):
...
+        self.nlp = English()
+        self.nlp.max_length = 1500000
+        self.nlp.add_pipe("entity_ruler").from_disk("rules/citation_patterns.jsonl")
+
+        self.postgresql = testing.postgresql.Postgresql()
+        self.db_conn = psycopg2.connect(**self.postgresql.dsn())
+        engine = create_engine(self.postgresql.url())
+
+        manifest_df = pd.read_csv("rules/2022_04_08_Citation_Manifest.csv")
+        manifest_df.to_sql("manifest", engine, if_exists="append", index=False)
```
anthonyhashemi, line 103: here we copy code from where we update the db to have the mainfest in one of the lambdas.

We need the manifest table as we access it in this these tests

tests/abbreviation_extraction_tests/test_abbreviations.py
```diff
@@ -1,218 +1,288 @@
+"""
```
anthonyhashemi, line 1: Ignore this file please, it has already been reviewed and approved in its own PR. I accidentally merged into this branch before this PR was merged in.

tests/caselaw_extraction_tests/test_case_citations.py
dragon-dxw: :+1:

tests/abbreviation_extraction_tests/test_abbreviations.py
dragon-dxw: Should be able to get rid of it from the commit by rebasing with `git fetch && git rebase -i origin/main`

dragon-dxw (approved):

dragon-dxw:
If you want a hand fixing the merge problem, let me know :)

tests/abbreviation_extraction_tests/test_abbreviations.py
anthonyhashemi: 👌

anthonyhashemi: ah that means I need to reopen the other PR first though - will do that

PR #164, opened 2023-03-01
Improve Github Actions

anthonyhashemi:
- Move all CI jobs that we want to run on every update to PR and main to one workflow yaml file.
- Split out jobs appropriately
- Remove redundancy between jobs 

Additional:
- removed unused workflows


.github/workflows/terraform_plan.yml
```diff
@@ -0,0 +1,53 @@
+name: terraform_plan
+
+on:
+  workflow_call:
+    secrets:
```
anthonyhashemi, line 5: I am passing secrets (https://docs.github.com/en/rest/actions/secrets?apiVersion=2022-11-28) from the parent workflow and that seems to work. If you think these should be some other type of variable (taging.caselaw.nationalarchives.gov.uk) or env and know how that works do let me know - I havent worked with github actions before.

.github/workflows/ci_lint_and_test.yml
```diff
@@ -11,7 +11,6 @@ on:
     branches:
       - main
```
anthonyhashemi, line 12: previously, the terraform validate and plan workflows that I have merged into this workflow did not run on merge to main, but I feel like there is no harm in doing that if we are going to run the tests and inits too. 

We do still have the apply.yml workflow that runs on merge to master which `terraform apply`s to staging and then production if that was successfull, but I dont think they conflict.

Let me know what you think

jacksonj04: This sounds solid to me.

.github/workflows/terraform_plan.yml
jacksonj04: This is the only bit I'm not 100% on, and that's not because I think this is wrong but because I don't know enough about how it was doing things before to say for sure that this is the right way (fsvo 'right').

The use of secrets is correct and I'm happy this isn't accidentally leaking tokens anywhere though!

jacksonj04 (approved):
Assuming that we know all of these secrets so we can update the names in GitHub to match the new ones, looks good as far as I can see. Some uncertainty at the deploy end, but at this stage I'm going to say that getting this pipe up and doing _something_ is an important outcome, and it looks right to me, so the best way to prove it is just to run it.

.github/workflows/ci_lint_and_test.yml
dragon-dxw: Only concern is if it takes ages, otherwise good.

.github/workflows/terraform_plan.yml
anthonyhashemi: it was the same for staging, prod up to the difference in these secrets

dragon-dxw (approved):
LGTM

anthonyhashemi:
> LGTM

it's pretty quick 😄

PR #165, opened 2023-03-02
Task/dry the deploy workflow

anthonyhashemi:
DRYed the CD workflow


anthonyhashemi:
Waiting for https://github.com/nationalarchives/ds-caselaw-data-enrichment-service/pull/166 to be merged in first so that the postgres upgrade issue that is causing terraform apply to fail on main is fixed.

dragon-dxw (approved):

PR #169, opened 2023-03-06
Fix the tidied up deploy workflow after realising error once merged to main

anthonyhashemi:
There was a bug in the deployment CICD workflow in PR: ta-enrichment-service/pull/165 which was only noticed when run on main.

This fixes it.


anthonyhashemi:
there was a rogue xml file in the initial PR - have removed and force pushed

dragon-dxw (approved):

PR #171, opened 2023-03-09
Bug/clean up oblique ref lambda/455

anthonyhashemi:
The bug https://trello.com/c/Xvk7Zg67/455-enrichment-duplication-bug is caused by the oblique_reference lambda. 

This first PR tidies that up first so that latter PRs are simpler

### DIFF
--- a/lambda/determine_oblique_references/index.py
+++ b/lambda/determine_oblique_references/index.py
@@ -1,16 +1,14 @@
 #!/usr/bin/env python3
 
-import csv
 import logging
 import os
-import re
 import urllib.parse
 
 import boto3
-from bs4 import BeautifulSoup
 
-from oblique_references.oblique_references import oblique_pipeline
-from replacer.second_stage_replacer import oblique_replacement
+from oblique_references.enrich_oblique_references import (
+    enrich_oblique_references,
+)
 
 LOGGER = logging.getLogger()
 LOGGER.setLevel(logging.INFO)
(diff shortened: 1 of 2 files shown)


oblique_references/enrich_oblique_references.py
```diff
@@ -0,0 +1,31 @@
+"""Module containing the function that enriches oblique references"""
+
+import re
+
+from oblique_references.oblique_references import oblique_pipeline
+from replacer.second_stage_replacer import oblique_replacement
+
+
+def enrich_oblique_references(file_content: str) -> str:
```
anthonyhashemi, line 9: have just copied and pasted the business logic from the lambda and moved into here

lambda/determine_oblique_references/index.py
```diff
@@ -1,16 +1,14 @@
 #!/usr/bin/env python3
 
-import csv
```
anthonyhashemi, line 3: unused

oblique_references/enrich_oblique_references.py
```diff
@@ -0,0 +1,31 @@
...
+    Returns:
+        (str): updated file content with enriched oblique references
+    """
+    # split file_content into header and judgment to ensure replacements only occur in judgment body
+    judgment_split = re.split("(</header>)", file_content)
+
+    resolved_refs = oblique_pipeline(judgment_split[2])
+
+    if not resolved_refs:
+        return []
```
dragon-dxw, line 25: this isn't a string, but the function apparently returns one.

dragon-dxw: and it doesn't resemble file content with enriched refs

oblique_references/enrich_oblique_references.py
```diff
@@ -0,0 +1,31 @@
...
+
+    output_file_data = oblique_replacement(judgment_split[2], resolved_refs)
+    # combine header with replaced text content before uploading to enriched bucket
+    judgment_split[2] = output_file_data
+
+    return "".join(judgment_split)
```
dragon-dxw, line 31: If you've split on header, I think you must join on header.

dragon-dxw (changes requested):
Looks good apart from the return values of the function being the wrong type, and empty?

Possible I've misunderstood some stuff.

oblique_references/enrich_oblique_references.py
anthonyhashemi: yeah tis the original and will be getting rid later

oblique_references/enrich_oblique_references.py
anthonyhashemi: as discussed on slack was a mistake - good catch

dragon-dxw (approved):

PR #173, opened 2023-03-10
Bug/refactor replacer/455

anthonyhashemi:
Another prep pr for https://trello.com/c/Xvk7Zg67/455-enrichment-duplication-bug:

Changes in this PR:

- Add successful unit test case for provision_replacement
- Remove unused import of provision_replacement
- Added typehints and addressed mypy complaints in second_stage_replacer.py

### DIFF
--- a/replacer/second_stage_replacer.py
+++ b/replacer/second_stage_replacer.py
@@ -7,15 +7,19 @@ Handles the replacements of oblique references and legislation provisions.
 
 import re
 from itertools import groupby
+from typing import Dict, List, Tuple, Union
 
 from bs4 import BeautifulSoup
 
+LegislationReference = Tuple[Tuple[int, int], str]
+LegislationReferenceReplacements = List[Dict[str, Union[str, int]]]
 
-def splitString(text, split_points):
+
+def split_string(text: str, split_points: List[int]) -> List:
     """
-    Splits a string at locations in the text where a reference was detected
-    :param text: XML file
-    :param split_points: list of positions of matches
+    Splits a string at given points in the text
+    :param text: some string of text
+    :param split_points: list of positions to split between
     :return: list of split strings
     """
     return list(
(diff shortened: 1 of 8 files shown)


replacer/second_stage_replacer.py
```diff
@@ -7,64 +7,94 @@
...
+    paragraphs = file_data.find_all("p")
+    ordered_reference_replacements = sorted(reference_replacements, key=key_func)
 
-    for p, matches in groupby(relevant_paras, key=key_func):
-        new_txt = BeautifulSoup(replacer(str(paras[p]), list(matches)), "lxml")
-        paras[p].replace_with(new_txt.p)
+    for paragraph_number, paragraph_reference_replacements in groupby(
+        ordered_reference_replacements, key=key_func
+    ):
+        paragraph_string = str(paragraphs[paragraph_number])
```
dragon-dxw, line 81: `key_func`has substantially changed code, feels like a different commit, as does `replace_references`.

This looks like it's part of the meat of the problem, rather than just useful tidying.

anthonyhashemi: do you mean provision_replacement/replace_references_by_paragraph has changed significantly?

it is just variable renaming and and breaking out one line logic into multiple lines to make everything easier to understand

happy to walk you through it

replacer/second_stage_replacer.py
```diff
@@ -7,64 +7,94 @@
...
+        for reference_replacement in reference_replacements
+        if isinstance((split_point := reference_replacement.get("ref_position")), int)
+    ]
+    split_text = split_string(text, split_points)
     enriched_text = text[: split_points[0]]
-    for spt, match in zip(split_text, detected_refs):
-        enriched_text += re.sub(re.escape(match["detected_ref"]), match["ref_tag"], spt)
+    for sub_text, reference_replacement in zip(split_text, reference_replacements):
+        detected_ref = reference_replacement["detected_ref"]
+        if not isinstance(detected_ref, str):
```
anthonyhashemi, line 48: this is just so we account for null situations

dragon-dxw (approved):

PR #174, opened 2023-03-13
Bug/enrichment dupication/455

anthonyhashemi:
Fixes bug https://trello.com/c/Xvk7Zg67/455-enrichment-duplication-bug

------
Bug description:
- certain documents, for example https://caselaw.nationalarchives.gov.uk/ewhc/ch/2023/257 were failing the validation at the end of the enrichment process and thus were not becoming enriched due to a mismatch of the text content of the judgment.
- From the validation error in the privilleged api we could see that para_18 was being duplicated due to a duplicate ref_para key.

------
Diagnosing the underlying Problem:

Thankfully at each stage of enrichment we store the partially enriched xml back to an s3 bucket, therefore we could see that between enrichment stage 1 completing and enrichment stage 2 completing (have a look at `docs/img/architecture.png` for reference) the duplicate paragraphs were being introduced. This meant that the `oblique_reference_replacer` lambda was the one that was causing the issue. 

Therefore I created a local script (which later became the basis for the integration test `test_enrich_oblique_references`) which would use the xml from stage_1 for this doc as the input and the stage_2 xml as the expected output of passing stage_1 through the oblique_ref enrichment. I then added breakpoints and investigated until a discrepancy was found.

The issue was that we were only considering the text contained inside paragraph tags when determining the position of references to replace but then were replacing on the whole xml content - thus replacing references in the wrong positions.

------
Solution:

The `determine_legislation_provisions` deals with `ref_para` and so I thought that it might be a good place to look for a solution. It similarly detects new refs that depend on the main legislations already having been enriched and so I attempted copying what it does, and use `replace_references_by_paragraph` and adjust the oblique ref detection logic to build up reference replacements which are relative per paragraph as opposed to all the paragraph content together.

That seemed to get the integration test passing! And so I stuck with that solution.

------
Tests:

Have added various unit tests on top of the integration test.

Do let me know if you think it is a problem to have real judgments as test fixtures

### DIFF
--- a/oblique_references/oblique_references.py
+++ b/oblique_references/oblique_references.py
@@ -1,6 +1,6 @@
 """
 @author: editha.nemsic
-This code handles the link of oblique references (i.e 'the Act' or 'the 1977 Act'). 
+This code handles the link of oblique references (i.e 'the Act' or 'the 1977 Act').
 
 This is done in the following way:
 1. We use the previously enriched judgment and identify where there are legislation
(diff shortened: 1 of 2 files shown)


dragon-dxw (approved):
✅

PR #176, opened 2023-03-14
Task/add test coverage/606

anthonyhashemi:
https://trello.com/c/Z8QT4CCd/606-add-test-coverage-to-enrichment-repo

This PR:

- adds `coverage.py` to the repo and updates CI and readme to use `coverage.py`
- deletes `tests/runner.py` and simplifies tests and updates docs to recommend to run tests with `python -m unittest`
- deletes old redundant tests
- fix and simplify old, previously nonrunning test suite

PR #177, opened 2023-03-14
More precommit and linting

dragon-dxw:
Based on top of https://github.com/nationalarchives/ds-caselaw-data-enrichment-service/pull/176 to have working tests -- adds more automated linting and manually cleans up some cruft


dragon-dxw:
There's a part of me that thinks we should not apply the E712/E711 changes because there might be brittleness due to numpy bools existing in other places.

setup.cfg
```diff
@@ -0,0 +1,42 @@
+[flake8]
+max-line-length = 4000
+extend-ignore = E203
+# E203: slices with [long thing : long thing] shouldn't have spaces
+exclude = .tox,.git,*/migrations/*,*/static/CACHE/*,docs,node_modules,venv
```
anthonyhashemi, line 5: why ignore migrations?

dragon-dxw: The simple reason is this is copypasted wholesale from the editor, and the migrations are typically autogenerated.
This isn't needed; overhaulled the config

anthonyhashemi (approved):

jacksonj04 (approved):

PR #179, opened 2023-03-15
More test fixes

dragon-dxw:
Discovered copy-pasting a test didn't run, so ran pytest to try to diagnose why and found some more tests.


lambdas
```diff
@@ -0,0 +1 @@
+lambda
```
dragon-dxw, line 1: I'm not a fan of this symlink, but I'm loathe to change the directory name without significant care and attention.

jacksonj04: 👍🏻

README.md
```diff
@@ -95,12 +95,14 @@ The standard mechanism for triggering the enrichment pipeline is via the TNA edi
...
+There is a suite of tests that can be run locally with
 
 ```sh
-python -m unittest
+pytest
 ```
 
+but you'll need to ensure you've installed `tests/requirements.txt`
+
 You can also obtain a test coverage report with `coverage run --source . -m unittest && coverage report`
```
anthonyhashemi, line 106: I think we should we just delete unittest to keep things simpler. I know that will mean adjusting a lot of the test classes that depend on TestCase or something so happy for that to be done later - I can make a ticket for it

jacksonj04 (changes requested):
Happy with all these changes, but could we also add an `include` statement in a `.coveragerc` file to make sure we're getting coverage on all the source files even if they're not hit by a test? The fact adding tests has reduced coverage makes me twitchy!

README.md
dragon-dxw: I've fixed this bit of the docs -- it's pytest in the CI.

I'm not convinced it's worth it -- if it gets in the way, sure, go for it; but I don't think it' something that has enough impact on us to justify the time.

That said, you disagree, so go for it... :)

dragon-dxw:
[More Test Fixes: Enrichment](https://trello.com/c/pgqIEVsk/665-more-test-fixes-enrichment)

anthonyhashemi (approved):

jacksonj04 (approved):

PR #192, opened 2023-03-21
Fix resolve overlap

dragon-dxw:
It looks like there's an issue where we're deleting things from a loop, and if there's three entries, we might delete one that then isn't there later, causing a crash.

### DIFF
--- a/src/legislation_extraction/legislation_matcher_hybrid.py
+++ b/src/legislation_extraction/legislation_matcher_hybrid.py
@@ -72,21 +72,43 @@ def resolve_overlap(results_dict):
     """
     qq = pd.DataFrame([results_dict])
     qq = qq.T.explode(0)[0].apply(pd.Series)
+
     qq.columns = keys
+
     # get refs that overlap in the text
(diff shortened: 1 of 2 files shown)


dragon-dxw:
[Legislation detection error for `ewca-civ-2023-263`](https://trello.com/c/2gojeBrt/618-legislation-detection-error-for-ewca-civ-2023-263)

src/legislation_extraction/legislation_matcher_hybrid.py
```diff
@@ -341,6 +363,8 @@ def leg_pipeline(leg_titles, nlp, docobj, conn):
 
     # merges the results of both matchers to return a single list of detected references
     results = mergedict(results[0], results[1])
+    print(results)
```
timcowlishaw, line 366: Is this a stray debug statement, or is STDOUT going somewhere important?

timcowlishaw (commented):
It's been a while since i've used pandas in anger for more than basic stuff, but this all looks very plausible to me - just one possible stray debug statement i found?

jacksonj04 (approved):

PR #193, opened 2023-03-22
Mypy precommit

dragon-dxw:
Add a precommit hook for mypy and silence errors. Some of these are significant issues which suggest the code would not have run (e.g. pass missing argument, no return value from a function leading to a codepath being impossible)

### DIFF
--- a/src/abbreviation_extraction/abbreviations.py
+++ b/src/abbreviation_extraction/abbreviations.py
@@ -334,9 +334,9 @@ class AbbreviationDetector:
         if matcher_output:
             matches_no_brackets = [(x[0], x[1] + 1, x[2] - 1) for x in matcher_output]
             filtered = filter_matches(matches_no_brackets, doc)
-            occurences = self.find_matches_for(filtered, doc)
+            occurrences = self.find_matches_for(filtered, doc)
 
-            for long_form, short_forms in occurences:
+            for long_form, short_forms in occurrences:
                 for short in short_forms:
                     short._.long_form = long_form
                     doc._.abbreviations.append(short)
--- a/src/abbreviation_extraction/abbreviations.py
+++ b/src/abbreviation_extraction/abbreviations.py
@@ -262,6 +262,11 @@ def verify_match_format(
         ):
             print(str(doc[start:end]))
             matcher_output.remove(match)
+    return None  # type:ignore
+    # This return value is depended on elsewhere in the code, but previously
+    # it did not exist at all. Understanding how the abbreviation code is
+    # meant to work is a task for the future.
+    # TODO: https://trello.com/c/horD3P3F/693-enrichment-pipeline-abbreviations-not-functional
 
 
 class AbbreviationDetector:
@@ -330,9 +335,13 @@ class AbbreviationDetector:
         matches_brackets = [(x[0], x[1], x[2]) for x in matches]
 
         matcher_output = verify_match_format(matches_brackets, doc)
+        # verify_match_format returns None, which is incorrect and means this section can not
+        # have been run in it's current format.
         print(matcher_output)
         if matcher_output:
-            matches_no_brackets = [(x[0], x[1] + 1, x[2] - 1) for x in matcher_output]
+            matches_no_brackets = [
+                (x[0], x[1] + 1, x[2] - 1) for x in matcher_output  # type:ignore
+            ]
             filtered = filter_matches(matches_no_brackets, doc)
             occurrences = self.find_matches_for(filtered, doc)
 
--- a/src/abbreviation_extraction/abbreviations.py
+++ b/src/abbreviation_extraction/abbreviations.py
@@ -11,7 +11,7 @@ ScispaCy repo here -> https://github.com/allenai/scispacy
 """
 
 from collections import defaultdict
-from typing import Dict, List, Optional, Set, Tuple
+from typing import Dict, Iterable, List, Optional, Set, Tuple
 
 from spacy.language import Language
 from spacy.matcher import Matcher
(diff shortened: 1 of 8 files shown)


src/tests/test_pipeline_components/test_replacer.py
```diff
@@ -56,7 +56,7 @@ def test_write_replacements_file(self):
             {"bar": ["[2022] 1 P&CR 123", "[2022] 1 P&CR 123", "2022"]},
         ]
 
-        for index, replacement in enumerate(replacements):
+        for index, replacement in enumerate(replacements_list):
```
dragon-dxw, line 59: Not sure if this could have worked. (Small chance it's code I already changed badly, but I think that was a different file)

src/abbreviation_extraction/abbreviations.py
```diff
@@ -229,7 +229,7 @@ def short_form_filter(span: Span) -> bool:
 
 def verify_match_format(
     matcher_output: List[Tuple[int, int, int]], doc: Doc
-) -> List[Tuple[Span, Span]]:
+) -> List[Tuple[int, int, int]]:
```
dragon-dxw, line 232: Not even sure where the Span output can come from, not in this code as far as I can see

src/abbreviation_extraction/abbreviations.py
```diff
@@ -262,6 +262,7 @@ def verify_match_format(
         ):
             print(str(doc[start:end]))
             matcher_output.remove(match)
+    return matcher_output
```
dragon-dxw, line 265: The absence of this return means an entire codepath in the calling function can't have been run -- this is a  fairly major change to how the program runs.

jacksonj04: I'm not happy with this, it makes me twitchy that the code path it was (in theory) supposed to have run has never been run, and presumably never tested.

Do we know what it was supposed to be doing?

dragon-dxw:
[Add mypy to enrichment repo](https://trello.com/c/vLDjdLlV/631-add-mypy-to-enrichment-repo)

src/abbreviation_extraction/abbreviations.py
dragon-dxw: The code appears to be removing matches which are too long (8+ words) and don't match the right shape (`("...")`). The tests do exercise this code path e.g. https://github.com/nationalarchives/ds-caselaw-data-enrichment-service/blob/957762fa0b06f84a66ca33272663965f1acae0a6/src/tests/abbreviation_extraction_tests/test_abbreviations.py#L207 during the self.nlp() step, but the doc they get back appears unchanged in both cases. The `filter_matches` function https://github.com/nationalarchives/ds-caselaw-data-enrichment-service/blob/957762fa0b06f84a66ca33272663965f1acae0a6/src/abbreviation_extraction/abbreviations.py#L114 seems to share much of the same logic and function correctly.

I'm moving towards "we delete this function entirely"

src/lambdas/vlex_upload/index.py
```diff
@@ -44,7 +44,7 @@ def process_event(sqs_rec):
         "Body"
     ].read()
 
-    upload_contents(file_content)
+    upload_contents(source_key, file_content)
```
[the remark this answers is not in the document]
anthonyhashemi: 😢

src/abbreviation_extraction/abbreviations.py
```diff
@@ -325,27 +290,13 @@ def __call__(self, doc: Doc) -> Doc:
...
-                for short in short_forms:
-                    short._.long_form = long_form
-                    doc._.abbreviations.append(short)
-
         return doc
 
     def find_matches_for(
         self, filtered: List[Tuple[Span, Span]], doc: Doc
     ) -> List[Tuple[Span, Set[Span]]]:
+        # This function is not exercised by tests and appears to be only called in .find()
```
anthonyhashemi, line 298: Do we want to keep this comment in the repo? If we are concerned with this not being used outside the module we could make it a private function `def _find_matches_for(

src/abbreviation_extraction/abbreviations.py
```diff
@@ -325,27 +290,13 @@ def __call__(self, doc: Doc) -> Doc:
         -------
         Doc, Doc object of the judgment content.
         """
-        matches = self.matcher(doc)
```
anthonyhashemi, line 328: why was this removed?

src/abbreviation_extraction/abbreviations.py
```diff
@@ -205,6 +205,7 @@ def filter_matches(
 
 
 def short_form_filter(span: Span) -> bool:
+    # This function is not exercised by tests, but is invoked by filter_matches
```
anthonyhashemi, line 208: Coverage is still very low, so still need to add lots of tests to cover everything, but we can make it a private method if we don't think it's worth testing on its own.

Likewise for `find_matches_for`

src/abbreviation_extraction/abbreviations.py
dragon-dxw: We've decided to flag up the issue and leave it mostly untouched for now.

src/abbreviation_extraction/abbreviations.py
dragon-dxw: We've decided against removing all this code, but agree we probably could remove it as it can't do anything.

src/legislation_provisions_extraction/legislation_provisions.py
```diff
@@ -15,10 +15,13 @@
 
 import os
 import re
+from typing import Any, Dict, List
 
 import numpy as np
 from bs4 import BeautifulSoup
 
+SectionDict = Dict[str, List[Any]]  # this is a guess
```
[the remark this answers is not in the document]
dragon-dxw: No.

anthonyhashemi (approved):
Approved assuming we have a bug ticket for the abbr pipeline

dragon-dxw:
Ticket added for Abbreviation Detector bug: https://trello.com/c/38Ee0vZD/704-abbreviation-pipeline-not-working

PR #194, opened 2023-03-24
Bug/dont catch lambda errors/640

anthonyhashemi:
We were previously catching all `Exception`s in `update_legislation_table` and `update_rules_processor`.

We noticed this when creating https://trello.com/c/RTIrUXGl/640-enrichment-legislation-table-not-been-updated-in-ages
since this `update_legislation_table` was erroring and showing so in the logs but the lambda function run was not failing as we did not reraise the exception after catching and logging like we do in all other lambda handler functions.

PR:
- Made sure to reraise caught exceptions in `update_legislation_table` and `update_rules_processor`
- Remove misleading postgres related info in exception
- Moved around some try except blocks in `update_rules_processor`

### DIFF
--- a/src/lambdas/update_legislation_table/index.py
+++ b/src/lambdas/update_legislation_table/index.py
@@ -176,5 +176,6 @@ def handler(event, context):
         engine.dispose()
         LOGGER.info("Legislation updated")
 
-    except (Exception, Error) as error:
-        LOGGER.error("Error while connecting to PostgreSQL: %s", error)
+    except Exception as exception:
+        LOGGER.error("Exception: %s", exception)
+        raise
--- a/src/lambdas/update_rules_processor/index.py
+++ b/src/lambdas/update_rules_processor/index.py
@@ -162,7 +162,6 @@ def lambda_handler(event, context):
         response = s3.get_object(Bucket=source_bucket, Key=source_key)
         csv_file = response["Body"].read().decode("utf-8")
         df = pd.read_csv(StringIO(csv_file))
-        # print(df)
 
         create_test_jsonl(source_bucket, df)
         jsonl_key = "test_citation_patterns.jsonl"
@@ -173,7 +172,11 @@ def lambda_handler(event, context):
 
         try:
             test_manifest(df, pattern_list)
-
+        except AssertionError:
+            LOGGER.error("Exception: Manifest test failed")
+            raise
+        
+        try:
             # write new jsonl file
             new_patterns_file = write_patterns_file(df["pattern"].to_list())
             upload_replacements(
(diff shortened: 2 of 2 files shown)


src/lambdas/update_legislation_table/index.py
```diff
@@ -176,5 +176,6 @@ def handler(event, context):
         engine.dispose()
         LOGGER.info("Legislation updated")
 
-    except (Exception, Error) as error:
```
anthonyhashemi, line 179: no need to catch Error, as Exception is the base exception so includes Error.

src/lambdas/update_rules_processor/index.py
```diff
@@ -173,29 +172,30 @@ def lambda_handler(event, context):
 
         try:
             test_manifest(df, pattern_list)
-
+        except AssertionError:
```
anthonyhashemi, line 175: the main try except was just trying to catch the assertion thrown from this test.

I'm not a fan of this whole test stuff, I'd prefer, a validate function which doesnt use asserts - but not for now

src/lambdas/update_rules_processor/index.py
```diff
@@ -173,29 +172,30 @@ def lambda_handler(event, context):
...
+            LOGGER.error("Exception: Manifest test failed")
```
anthonyhashemi, line 176: thinking about it - this exception is kind of vague. I don't yet understand what it means as Im not sure what the purpose of the test is (havent looked at the test yet)

timcowlishaw (approved):
Does what it says on the tin! Agree that using exception flow for that manifest check seems a bit suspect, but also that that can be a later improvement :-)

src/lambdas/update_legislation_table/index.py
```diff
@@ -176,5 +176,6 @@ def handler(event, context):
...
-        LOGGER.error("Error while connecting to PostgreSQL: %s", error)
```
anthonyhashemi, line 180: remove this Postgres message since it is misleading as we also try to connect to the legislation endpoint and are just doing much more than connecting to postgresql.

This confused me when we had actually the wrong legislation api creds - I was thinking we had the wrong postgresql creds!

src/lambdas/update_rules_processor/index.py
dragon-dxw: Not sure if it was you or Nick who mentioned https://snyk.io/blog/the-dangers-of-assert-in-python/

(asserts can be silently ignored in "production" environments!)

dragon-dxw (approved):

src/lambdas/update_rules_processor/index.py
anthonyhashemi: Ah okay, a message like that would be a bit clearer

I will make a ticket to clean this up and change it to be a validation function and fix up any error message there

anthonyhashemi:
Ticket for addressing the whole test_manifest and assertions in source code stuff discussed in the PR: https://trello.com/c/GHWO6tUo/694-enrichment-code-improvement-convert-test-logic-to-validation-check-in-updaterulesmanifest-lambda

PR #196, opened 2023-04-11
Simplify db env vars

anthonyhashemi:
This PR:

- Reduces database env var reading boiler plate
- make database env var names more explicit (without prefixing DATABASE_ to PASSWORD to make DATABASE_PASSWORD, sourcing a .env locally would mess up the shell so this was necessary as I added that new env as discussed in comments)
- and remove unused TABLE_NAME env var (discussed in comments)

### DIFF
--- a/src/database/db_connection.py
+++ b/src/database/db_connection.py
@@ -26,7 +26,6 @@ def create_connection(db, user, password, host, port):
     except (Exception, psycopg2.DatabaseError) as error:
         print(error)
         raise
-    # return conn
 
 
 def get_manifest_row(conn, rule_id):
--- a/src/database/db_connection.py
+++ b/src/database/db_connection.py
@@ -5,7 +5,7 @@ import pandas as pd
 import psycopg2
 
 
-def create_connection(db, user, password, host, port):
+def create_connection(db, user, password, host, port) -> psycopg2.extensions.connection:
     """
     Connect to the PostgreSQL database server
     :param db
--- a/src/utils/environment_helpers.py
+++ b/src/utils/environment_helpers.py
@@ -25,9 +25,11 @@ def validate_env_variable(env_var_name):
 
 
 def get_database_password():
-    password = validate_env_variable("DATABASE_PASSWORD")
-    if password is not None:
+    try:
+        password = validate_env_variable("DATABASE_PASSWORD")
         return password
+    except Exception:
+        pass
     aws_secret_name = validate_env_variable("SECRET_PASSWORD_LOOKUP")
     aws_region_name = validate_env_variable("REGION_NAME")
     return _get_aws_secret(aws_secret_name, aws_region_name)
--- a/src/utils/environment_helpers.py
+++ b/src/utils/environment_helpers.py
@@ -24,13 +24,7 @@ def validate_env_variable(env_var_name):
     return env_variable
 
 
-def get_database_password():
-    aws_secret_name = validate_env_variable("SECRET_PASSWORD_LOOKUP")
-    aws_region_name = validate_env_variable("REGION_NAME")
-    return _get_aws_secret(aws_secret_name, aws_region_name)
-
-
-def _get_aws_secret(aws_secret_name, aws_region_name):
+def get_aws_secret(aws_secret_name, aws_region_name):
     """
     Get aws secret
     """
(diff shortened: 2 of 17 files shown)


src/database/db_connection.py
```diff
@@ -3,6 +3,38 @@
...
+)
+
+DATABASE_NAME = validate_env_variable("DATABASE_NAME")
+USERNAME = validate_env_variable("DATABASE_USERNAME")
+HOST = validate_env_variable("DATABASE_HOSTNAME")
+PORT = validate_env_variable("DATABASE_PORT")
+PASSWORD = get_database_password()
+
+
+def init_db_engine() -> sqlalchemy.Engine:
```
anthonyhashemi, line 20: we were repeating these 2 db creation bits of logic so moved into reusable functions in db_connection file. Also moved the reading of the db env vars here too as constants that can be used in both.

At a later point I think we should probably align all lambdas to just use one of these 2 methods of db connection and get rid of the other - as it confuses/bloats the code unnecessarily

terraform/modules/lambda_s3/lambda.tf
```diff
@@ -1108,16 +1105,11 @@ module "lambda-update-rules-processor" {
...
-    PORT                   = "5432"
+    DATABASE_USERNAME      = "root"
+    DATABASE_HOSTNAME      = "${var.postgress_hostname}"
+    DATABASE_PORT          = "5432"
     SECRET_PASSWORD_LOOKUP = "${var.postgress_master_password_secret_id}"
     REGION_NAME            = "${local.region}"
-    HOSTNAME               = "${var.postgress_hostname}"
-    # SPARQL_USERNAME = "${var.sparql_username}"
-    # SPARQL_PASSWORD = "${var.sparql_password}"
-    SPARQL_USERNAME = ""
```
anthonyhashemi, line 1119: unused env vars

terraform/modules/lambda_s3/lambda.tf
```diff
@@ -300,12 +300,11 @@ module "lambda-determine-replacements-caselaw" {
 
   environment_variables = {
     DATABASE_NAME          = "rules"
-    TABLE_NAME             = "rules"
```
anthonyhashemi, line 303: table_name was incorrect and unused in all lambdas (we dont use a rules table anywhere) - it seems more specific to the lambdas as opposed to the db we are connecting to - so instead of adjusting the table_name just removed the env var and continued to expliilty specify it in the lambda logic

src/utils/environment_helpers.py
```diff
@@ -0,0 +1,66 @@
+import base64
+import logging
+import os
+
+import boto3
+
+LOGGER = logging.getLogger()
+LOGGER.setLevel(logging.INFO)
+
+
```
anthonyhashemi, line 10: `validate_env_variable` was repeated so many times.

as was getLoginSecrets class which was unnecesarily a class, which I have made `_get_aws_secret` here.


Have created `get_database_password` which takes the repeated code of getting the aws secrets as before but now also looks to see if there is a password specified, as otherwise it wont work local dbs when doing local dev without a aws setup

src/utils/environment_helpers.py
```diff
@@ -0,0 +1,70 @@
...
+    return env_variable
+
+
+def get_database_password():
+    try:
+        password = validate_env_variable("DATABASE_PASSWORD")
+        return password
+    except Exception:
+        pass
```
dragon-dxw, line 32: Don't like this -- we're silently eating all exceptions from the db password being absent?

Can we either:
1) create a named exception and test for that, or
2) explicitly check if DATABASE_PASSWORD is empty/absent and continue if it is?

dragon-dxw (approved):

src/utils/initialise_db.py
```diff
@@ -0,0 +1,39 @@
...
+    validate_env_variable,
+)
+
+DATABASE_NAME = validate_env_variable("DATABASE_NAME")
+USERNAME = validate_env_variable("DATABASE_USERNAME")
+HOST = validate_env_variable("DATABASE_HOSTNAME")
+PORT = validate_env_variable("DATABASE_PORT")
+PASSWORD = get_database_password()
+
+
```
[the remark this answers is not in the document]
dragon-dxw: I'm for refactoring this.

src/utils/environment_helpers.py
anthonyhashemi: Have just removed DATABASE_PASSWORD and its associated logic and test. Only added for tests but we can and do already get around it with moto's mock_secretsmanager

PR #197, opened 2023-04-12
Refactor update_legislation_table lambda and add tests for it

anthonyhashemi:
https://trello.com/c/RTIrUXGl/640-enrichment-legislation-table-not-been-updated-in-ages

This PR:

- refactors the update_legislation_table lambda handler function
- adds tests for it and extracted functions

### DIFF
--- a/.github/workflows/ci_lint_and_test.yml
+++ b/.github/workflows/ci_lint_and_test.yml
@@ -53,6 +53,9 @@ jobs:
           pip3 install -r src/tests/requirements.txt
 
       - name: Run python tests with coverage
+        env:
+          SPARQL_USERNAME: ${{ secrets.SPARQL_USERNAME }}
+          SPARQL_PASSWORD: ${{ secrets.SPARQL_PASSWORD }}
         id: run_python_tests
         run: coverage run --source . -m pytest
 
--- a/.github/workflows/ci_lint_and_test.yml
+++ b/.github/workflows/ci_lint_and_test.yml
@@ -104,8 +104,8 @@ jobs:
       - name: Run Terraform Validate
         id: run_terraform_validate
         run: |
-          set +e
-          set +o pipefail
+          set -e
+          set -o pipefail
           cd terraform
           terraform init -backend=false
           terraform validate -no-color
--- a/src/lambdas/update_legislation_table/fetch_legislation.py
+++ b/src/lambdas/update_legislation_table/fetch_legislation.py
@@ -2,6 +2,7 @@ import datetime
 import logging
 import re
 from io import BytesIO
+from typing import Optional
 
 import pandas as pd
 from SPARQLWrapper import CSV, SPARQLWrapper
@@ -13,7 +14,7 @@ LEGISLATION_API_URL = "https://www.legislation.gov.uk/sparql"
 
 
 def fetch_legislation(
-    sparql_username: str, sparql_password: str, days: int
+    sparql_username: str, sparql_password: str, days: Optional[int]
 ) -> pd.DataFrame:
     """
     Fetch new legislation from legislation.gov.uk since the start day, if given,
--- a/src/lambdas/update_legislation_table/fetch_legislation.py
+++ b/src/lambdas/update_legislation_table/fetch_legislation.py
@@ -1,0 +1,80 @@
+import datetime
+import logging
+import re
+from io import BytesIO
+
+import pandas as pd
+from SPARQLWrapper import CSV, SPARQLWrapper
+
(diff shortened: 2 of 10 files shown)


.github/workflows/ci_lint_and_test.yml
```diff
@@ -53,6 +53,9 @@ jobs:
           pip3 install -r src/tests/requirements.txt
 
       - name: Run python tests with coverage
+        env:
+          SPARQL_USERNAME: ${{ secrets.SPARQL_USERNAME }}
```
anthonyhashemi, line 57: have added these to github actions as secrets and the corresponding tests that rely on these do pass :)

src/lambdas/update_legislation_table/fetch_legislation.py
```diff
@@ -0,0 +1,81 @@
...
+
+    results = sparql.query().convert()
+    df = pd.read_csv(BytesIO(results))
+    LOGGER.info("Legislation retrieved")
+
+    df = _enhance_legislation_data(df)
+    return df
+
+
+def _enhance_legislation_data(df):
```
anthonyhashemi, line 73: I'd probably like to break this out into its own public function but I think I have spent enough time on this for now 😄

src/lambdas/update_legislation_table/index.py
```diff
@@ -1,103 +1,56 @@
...
-    df = df[~df["candidate_titles"].isna()].drop_duplicates("candidate_titles")
-    df["for_fuzzy"] = df.candidate_titles.apply(
-        lambda x: re.search(r"Act\s+(\d{4})", x) != None
-    )
-    return df
+    try:
+        trigger_date = event.get("trigger_date")
+        if not isinstance(trigger_date, int):
+            trigger_date = None
+        update_legislation_table(trigger_date)
```
anthonyhashemi, line 24: This way if we ever want to migrate from aws, things should be easier

src/lambdas/update_legislation_table/tests/test_fetch_legislation.py
```diff
@@ -0,0 +1,94 @@
...
+    del os.environ["SPARQL_USERNAME"]
+    del os.environ["SPARQL_PASSWORD"]
+
+
+frozen_date = datetime.datetime(2023, 4, 13)
+
+
+@pytest.mark.integration
+@patch("lambdas.update_legislation_table.fetch_legislation.datetime.datetime")
+def test_fetch_legislation_integration(mock_datetime, set_env_vars) -> None:
```
anthonyhashemi, line 37: I wanted to make sure we had some test that actually hit the legislation api to make sure our query was correct - alternatively I could have just made sure the query was right by checking in the api and then trusting that as long as we send the write string to a mocked api we're good - but this gives me more confidence. e.g. I had made a typo and the query was actually broken... but on the other hand if the test was written before I had done that it would have been caught by the mocked test as well.

src/lambdas/update_legislation_table/tests/test_update_legislation.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@patch("lambdas.update_legislation_table.index.fetch_legislation")
+def test_update_legislation_table(
+    mock_fetch_legislation,
+    monkeypatch,
+    setup_moto_secrets_manager,
+    test_db_connection,
+):
+    """
```
anthonyhashemi, line 53: this is the test of the main function  (ignoring the handler which is tested simply in the other file)

I have mocked fetch_legislation here to keep things simple.

The fact that fetch_legislation is tested fully with access to the db, in cobjunction with this test gives me confidence in this lambda function now.

Alternatively, I could do the full integration test at this level also but then I feel like we'd be calling the api twice unnecessarily

jacksonj04 (approved):
Lovely bit of work, really reassuring to see these tests in place. Nice bit of tidying around the logger too.

src/lambdas/update_legislation_table/index.py
```diff
@@ -1,103 +1,40 @@
...
+from lambdas.update_legislation_table.fetch_legislation import (
+    fetch_legislation,
+)
 from utils.environment_helpers import validate_env_variable
 from utils.initialise_db import init_db_engine
 
 LOGGER = logging.getLogger()
 LOGGER.setLevel(logging.INFO)
 
+LEGISLATION_TABLE_NAME = "ukpga_lookup"
```
jacksonj04, line 13: Happy with this not being an env variable, it seems pretty stable and if we want to change it we're going to have to do a fair bit in all the places for no real benefit.

PR #198, opened 2023-04-12
remove_duplicates as part of update_legislation_table lambda

anthonyhashemi:
https://trello.com/c/RTIrUXGl/640-enrichment-legislation-table-not-been-updated-in-ages

The enrichment service has a db which just mirrors the legislation api's db (for efficiency reasons.

Without this PR, the lambda used to keep it up to date `update_legislation_table` _could_ add duplicate entries to the db if the trigger_date went back to a date already covered by the db.

This PR addresses that and removes any duplicate entries at the end of the `update_legislation_table` lambda.

### DIFF
--- a/src/lambdas/update_legislation_table/test_update_legislation.py
+++ b/src/lambdas/update_legislation_table/test_update_legislation.py
@@ -21,9 +21,26 @@ postgresql_my = factories.postgresql("postgresql_my_proc")
 def test_db_connection(postgresql_my):
     conn = postgresql_my
     conn.autocommit = True
-    conn.cursor().execute("CREATE DATABASE ukpga_lookup")
+    sql_query = """
+    CREATE TABLE ukpga_lookup (
+        ref VARCHAR(100) NOT NULL,
+        title VARCHAR(100) NOT NULL,
+        ref_version VARCHAR(100) NOT NULL,
+        shorttitle VARCHAR(100) NOT NULL,
+        citation VARCHAR(100) NOT NULL,
+        acronymcitation VARCHAR(100) NOT NULL,
+        year BIGINT NOT NULL,
+        candidate_titles VARCHAR(100) NOT NULL,
+        for_fuzzy BOOLEAN NOT NULL
+    );
+    INSERT INTO ukpga_lookup (ref, title, ref_version, shorttitle, citation, acronymcitation, year, candidate_titles, for_fuzzy)
+    VALUES
+        ('a', 'a_title', 'a_ref_version', 'a_shorttitle', 'a_citation', 'a_acronymcitation', 2001, 'a_candidate_titles', true),
+        ('b', 'b_title', 'b_ref_version', 'b_shorttitle', 'b_citation', 'b_acronymcitation', 2001, 'b_candidate_titles', true)
+    """
+    conn.cursor().execute(sql_query)
     yield conn
-    conn.cursor().execute("DROP DATABASE ukpga_lookup")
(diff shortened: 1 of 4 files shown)


src/lambdas/update_legislation_table/test_update_legislation.py
```diff
@@ -21,9 +21,26 @@
...
+        for_fuzzy BOOLEAN NOT NULL
+    );
+    INSERT INTO ukpga_lookup (ref, title, ref_version, shorttitle, citation, acronymcitation, year, candidate_titles, for_fuzzy)
+    VALUES
+        ('a', 'a_title', 'a_ref_version', 'a_shorttitle', 'a_citation', 'a_acronymcitation', 2001, 'a_candidate_titles', true),
+        ('b', 'b_title', 'b_ref_version', 'b_shorttitle', 'b_citation', 'b_acronymcitation', 2001, 'b_candidate_titles', true)
+    """
+    conn.cursor().execute(sql_query)
     yield conn
-    conn.cursor().execute("DROP DATABASE ukpga_lookup")
```
dragon-dxw, line 26: Was the database named this (and thus it dropped all the data) or was this failing?
Can we catch and raise errors from the SQL?
(Or does that happen automatically?)

src/lambdas/update_legislation_table/update_legislation_table/database.py
```diff
@@ -0,0 +1,35 @@
+"""
+This module contains utility functions for working with databases using SQLAlchemy.
+"""
+
+from sqlalchemy import Connection, text
+
+
+def remove_duplicates(db_conn: Connection, table_name: str):
```
dragon-dxw, line 8: There's a part of me wondering whether constraints on the db is a more sensible approach -- we change the schema of the db to require a UNIQUE ctid and ON CONFLICT DO NOTHING?

dragon-dxw (approved):

src/lambdas/update_legislation_table/update_legislation_table/database.py
anthonyhashemi: Ah, I didn't even consider that - good point - I did think this is some logic that I shouldn't need to be creating.

Do you know how we would go about that?

src/lambdas/update_legislation_table/test_update_legislation.py
anthonyhashemi: tbh neither, I was just being consistent with what we are doing elsewhere. I dont think it matters too much as the test postgres database gets destroyed at the end of the test

src/lambdas/update_legislation_table/update_legislation_table/database.py
anthonyhashemi: I did look into this suggestion, even adding the contraint with the query:

`ALTER TABLE ukpga_lookup ADD CONSTRAINT ukpga_lookup_unique_rows_key UNIQUE (ref, title, ref_version, shorttitle, citation, acronymcitation, year, candidate_titles, for_fuzzy);`


But then I realised we don't have a way to specify `ON CONFLICT DO NOTHING` at the moment due to using df.to_sql which doesnt seem to have a parameter to do that.

We could of course rewrite the way we write to the db but I think its not worth the effort right now given that I have already written the duplicate function.

What do you think?

anthonyhashemi: @jacksonj04 thoughts?

PR #200, opened 2023-04-13
Fix issues with github actions for deployment

dragon-dxw:
* we should delete the "On PRs, deploy to staging" action before merging.
* ~~there may be other parts of this where we `set -e` but they're for a different PR.~~
* we also get rid of a number of places we were using `set +e` where it wasn't raising an error.
* I'm not confident I'm doing the right thing with the jsonl file, but not sure what else would be right.

### DIFF
--- a/.github/workflows/deploy_single_environment.yml
+++ b/.github/workflows/deploy_single_environment.yml
@@ -41,9 +41,9 @@ jobs:
 
       - name: Assemble Non-Dockerised Lambda
         run: |
-          set +e
-          set +o pipefail
-          
+          set -e
+          set -o pipefail
+
           cd src/
 
           mkdir lambdas/determine_replacements_caselaw/caselaw_extraction
@@ -108,11 +108,11 @@ jobs:
       # Pushing images to ECR last
       - name: Build Docker Images For Lambda
         run: |
-          set +e
-          set +o pipefail
-          
-          aws ecr get-login-password --region eu-west-2 | docker login --username AWS --password-stdin ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com          
-          
+          set -e
+          set -o pipefail
+
+          aws ecr get-login-password --region eu-west-2 | docker login --username AWS --password-stdin ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com
+
           cd "src/lambdas/determine_replacements_caselaw"
           docker buildx build --load -t tna-s3-tna-ecr-repository-{{ secrets.app_env }} .
           docker tag tna-s3-tna-ecr-repository-{{ secrets.app_env }}:latest ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com/tna-s3-tna-ecr-repository-{{ secrets.app_env }}:latest
--- a/.github/workflows/deploy_single_environment.yml
+++ b/.github/workflows/deploy_single_environment.yml
@@ -117,61 +117,62 @@ jobs:
           aws ecr get-login-password --region eu-west-2 | docker login --username AWS --password-stdin ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com
 
           cd "src/lambdas/determine_replacements_caselaw"
-          docker buildx build --load -t tna-s3-tna-ecr-repository-{{ secrets.app_env }} .
-          docker tag tna-s3-tna-ecr-repository-{{ secrets.app_env }}:latest ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com/tna-s3-tna-ecr-repository-{{ secrets.app_env }}:latest
-          docker push ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com/tna-s3-tna-ecr-repository-{{ secrets.app_env }}:latest
+          docker buildx build --load -t tna-s3-tna-ecr-repository-${{ secrets.app_env }} .
+          docker tag tna-s3-tna-ecr-repository-${{ secrets.app_env }}:latest ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com/tna-s3-tna-ecr-repository-${{ secrets.app_env }}:latest
+          docker push ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com/tna-s3-tna-ecr-repository-${{ secrets.app_env }}:latest
+
           cd "${GITHUB_WORKSPACE}"
 
           cd "src/lambdas/determine_replacements_legislation"
-          docker buildx build --load -t tna-s3-tna-ecr-repository-legislation-{{ secrets.app_env }} .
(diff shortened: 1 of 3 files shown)


.github/workflows/cd_deploy_only_staging.yml
```diff
@@ -0,0 +1,29 @@
+name: cd_deploy_only_staging
```
anthonyhashemi, line 1: So this was just so you could test stuff without merging in, got it 👍

.github/workflows/deploy_single_environment.yml
```diff
@@ -108,67 +111,69 @@ jobs:
...
+          docker push ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com/tna-s3-tna-ecr-repository-abbreviations-${{ secrets.app_env }}:latest
           cd "${GITHUB_WORKSPACE}"
 
           cd "src/lambdas/update_legislation_table"
-          docker buildx build --load -t tna-s3-tna-ecr-repository-legislation-update-{{ secrets.app_env }} .
-          docker tag tna-s3-tna-ecr-repository-legislation-update-{{ secrets.app_env }}:latest ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com/tna-s3-tna-ecr-repository-legislation-update-{{ secrets.app_env }}:latest
-          docker push ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com/tna-s3-tna-ecr-repository-legislation-update-{{ secrets.app_env }}:latest
+          docker buildx build --load -t tna-s3-tna-ecr-repository-legislation-update-${{ secrets.app_env }} .
+          docker tag tna-s3-tna-ecr-repository-legislation-update-${{ secrets.app_env }}:latest ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com/tna-s3-tna-ecr-repository-legislation-update-${{ secrets.app_env }}:latest
```
anthonyhashemi, line 141: so was the main issue that the $ sign was missing and so the variable names were being used as opposed to the variables themselves and thus the lambdas werent being updated as they pointed to other docker images?

.github/workflows/deploy_single_environment.yml
```diff
@@ -108,67 +111,69 @@ jobs:
...
-          set +o pipefail
-          
-          aws ecr get-login-password --region eu-west-2 | docker login --username AWS --password-stdin ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com          
-          
+          set -e
+          set -o pipefail
+
+          aws ecr get-login-password --region eu-west-2 | docker login --username AWS --password-stdin ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com
+
+          echo docker buildx build . --load -t tna-s3-tna-ecr-repository-${{ secrets.app_env }}
```
anthonyhashemi, line 119: what's this new echo line doing?

.github/workflows/cd_deploy_only_staging.yml
dragon-dxw: 💯

.github/workflows/deploy_single_environment.yml
```diff
@@ -108,67 +111,69 @@ jobs:
       # Pushing images to ECR last
       - name: Build Docker Images For Lambda
         run: |
-          set +e
-          set +o pipefail
-          
-          aws ecr get-login-password --region eu-west-2 | docker login --username AWS --password-stdin ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com          
-          
+          set -e
```
anthonyhashemi, line 114: yeah this makes sense to exit on first error now that the repo is more stable. 

I can see why they may have wanted it to continue at the start though if there were lots of errors

.github/workflows/deploy_single_environment.yml
```diff
@@ -41,17 +41,20 @@ jobs:
...
           mkdir lambdas/determine_replacements_caselaw/caselaw_extraction
           cp -r caselaw_extraction/*.py lambdas/determine_replacements_caselaw/caselaw_extraction/
           cp caselaw_extraction/requirements.txt lambdas/determine_replacements_caselaw/requirements.txt
           mkdir lambdas/determine_replacements_caselaw/rules
           # cp caselaw_extraction/rules/citation_patterns.jsonl lambdas/determine_replacements_caselaw/rules/citation_patterns.jsonl
-          cp citation_patterns.jsonl lambdas/determine_replacements_caselaw/
+          # cp citation_patterns.jsonl lambdas/determine_replacements_caselaw/
+          # dragon: the first of these was commented out, but could copy a file. The second cannot...
+          # but we need a citations_patterns.json  in the root lambda directory, so let's make a chimera.
+          cp caselaw_extraction/rules/citation_patterns.jsonl lambdas/determine_replacements_caselaw/
```
anthonyhashemi, line 57: The only place in python source files that reference the exact string `citation_patterns.jsonl` (excluding commented code) is:

And it is also referenced in:

- line 311: terraform/modules/lambda_s3/lambda.tf : `RULES_FILE_KEY      = "citation_patterns.jsonl"`
      -  `RULES_FILE_KEY` is referenced only in `src/lambdas/determine_replacements_caselaw/index.py` for determining the rules for caselaw pattern matching.
      
`src/caselaw_extraction/rules/citation_patterns.jsonl` is the only `citation_patterns.jsonl` file in the repo anymore. so yeah `cp caselaw_extraction/rules/citation_patterns.jsonl lambdas/determine_replacements_caselaw/` seems right to me.

Was the line has been updated working - I would have thought not?

And to clarify, this is a separate issue to the function not updating given the code change though, right?

dragon-dxw: is the comment useful? Should I just have the version I've crafted?

dragon-dxw:
Adding a commit to set -e all the things; tested on https://github.com/nationalarchives/ds-caselaw-data-enrichment-service/pull/201

.github/workflows/deploy_single_environment.yml
anthonyhashemi: I _think_ it is unneeded since in the dockerfile we copy the whole caselaw_extraction/ folder 

`COPY caselaw_extraction/ ${LAMBDA_TASK_ROOT}/caselaw_extraction/`

and so should include `src/caselaw_extraction/rules/citation_patterns.jsonl`

I think previously there was a citation_patterns.jsonl at the root of the repo, maybe because it was used in multiple lambdas and so they had to copy it into both workspaces.

Ideally we wouldnt need to do all this manual copying and use shared lambda layers for any shared code. My WIP PR which I put on hold a while ago because it wasnt the most pressing thing, is working towards that: https://github.com/nationalarchives/ds-caselaw-data-enrichment-service/pull/184

dragon-dxw:
[Certain enrichment lambda functions not updating on merge even though terraform updating env variables (causing errors)](https://trello.com/c/lLABy4j9/791-certain-enrichment-lambda-functions-not-updating-on-merge-even-though-terraform-updating-env-variables-causing-errors)

PR #202, opened 2023-04-14
Fix package imports in update_legislation_table lambda

anthonyhashemi:
Fix package imports in update_legislation_table lambda:

I was getting errors due to the refactor since the lambda function didnt have utils module in its environment - so for that I copied it in to the env like we do for the other zip file lambdas,

AND 

the `src/lambdas/update_legislation_table/fetch_legislation.py` file was also not being found due to me trying to import it from `lamdbas.update_legislation_table.fetch_legislation import fetch_...` as opposed to `from fetch_legislation import ...`

The whole lambda folder structure and env setup is a bit complicated atm, but hopefully when https://github.com/nationalarchives/ds-caselaw-data-enrichment-service/pull/184 is complete - things will be easier (by using lambda layers for shared code and having dev/test dependencies per lambda as opposed to the generic test requirements.txt we have atm)


jacksonj04 (approved):
Thought: Given this has a bunch of formatting changes too, is it worth just running a formatter over the whole codebase in one shot as a separate PR, so they're clearer in the future?

anthonyhashemi:
> Thought: Given this has a bunch of formatting changes too, is it worth just running a formatter over the whole codebase in one shot as a separate PR, so they're clearer in the future?

I don't think there are any formatting changes here. Unless I have something odd going on with my diff? What changes are you seeing?

PR #204, opened 2023-04-18
Bug/final fixes for update legislation table lambda/680

anthonyhashemi:
Have fixed the dockerfile (as well as simplifying it) and added a dockerignore so that we do not include test files, cache files and .env files.

Also had to the fixe the version of sqlalchemy being installed in this dockerfile as we now import Connection object in the remove_duplicates file.

Have now tested the docker image locally and shown it works.

Screenshot 2023-04-18 at 12 57 38


src/lambdas/update_legislation_table/Dockerfile
```diff
@@ -1,32 +1,14 @@
...
-# output stream, which means that logs can be delivered to the user quickly.
-# PYTHONDONTWRITEBYTECODE keeps Python from writing the .pyc files, which are unnecessary in this case
 ENV PYTHONDONTWRITEBYTECODE=1 PYTHONUNBUFFERED=1 PYTHONIOENCODING=UTF-8
 
 
 RUN yum update -y && \
     rm -Rf /var/cache/yum
 
-COPY requirements.txt ${LAMBDA_TASK_ROOT}
+COPY . ${LAMBDA_TASK_ROOT}
```
dragon-dxw, line 9: I am somewhat in favour of saying

copy requirements
run pip
copy the rest

so that pip gets cached

but not essential.

dragon-dxw (approved):

src/lambdas/update_legislation_table/Dockerfile
anthonyhashemi: ah great point - updating 👍

PR #207, opened 2023-04-21
Upgrade pandas to 1.5.3

anthonyhashemi:
We had updated sqlalchemy to 2.0.0 but only newer versions of pandas df.to_sql is compatible with it, so upgrading pandas  to 1.5.3.
- again not caught by unit tests in ci due to how we have the test env/lambda env set up - ticket to address this exists.

more info on this here: https://stackoverflow.com/questions/75273632/got-multiple-values-for-argument-schema

### DIFF
--- a/src/lambdas/update_legislation_table/requirements.txt
+++ b/src/lambdas/update_legislation_table/requirements.txt
@@ -1,4 +1,4 @@
 psycopg2-binary==2.9.1
-pandas==1.1.5
+pandas==1.5.3
 sqlalchemy==2.0.4
 SPARQLWrapper==2.0.0


jacksonj04 (approved):

jacksonj04:
Do we want to set up Dependabot to just automatically pull versions up, or is that too much of a dangerous game until we get testing solved?

anthonyhashemi:
> Do we want to set up Dependabot to just automatically pull versions up, or is that too much of a dangerous game until we get testing solved?

I think it is too dangerous until we have fixed the test env issue but definitely worth it once done 👍

PR #213, opened 2023-06-13
Remove comment about just listing S3 buckets

dragon-dxw:
Does seem to do a lot more, these days.


anthonyhashemi:
> I'm happy with the changes, but I'm not sure about those test failures!

I have forced rerun - rhere was a bug with spacy I have fixed in my PR

PR #219, opened 2023-06-15
Reduce duplication in queue tf file

anthonyhashemi:
https://trello.com/c/NrZGHaJg/997-some-enrichment-lambda-functions-arent-alerting-us-of-errors

Breaking Changes:

- all the topics into one new topic called `Enrichment Error` as opposed to specific ones like `Fetch XML Error`, but the underlying errors alerts for each lambda function will still exist and are distinguishable within each topic.

Refactor-Only Changes:

- All lambda functions now have their corresponding error `aws_cloudwatch_metric_alarm` created  dynamically rather than lots of duplicated boilerplate - utilising the `count` meta-argument https://developer.hashicorp.com/terraform/language/meta-arguments/count
- Subscriptions to the new topic has less boilerplate by using a email list variable.

### DIFF
--- a/terraform/modules/lambda_s3/variables.tf
+++ b/terraform/modules/lambda_s3/variables.tf
@@ -1,3 +1,9 @@
+# Create SNS topic subscription for each developer
+variable "developer_emails" {
+  type    = list(string)
+  default = ["anthony.hashemi@nationalarchives.gov.uk", "david.mckee@dxw.com"]
+}
+
 variable "aws_profile" {
   type    = string
   default = "default"
--- a/terraform/modules/lambda_s3/queue.tf
+++ b/terraform/modules/lambda_s3/queue.tf
@@ -329,69 +329,11 @@ resource "aws_sns_topic" "validation_updates_error" {
   name = "validation-updates-error-topic"
 }
 
-resource "aws_sns_topic" "db_backup_error" {
-  name         = "db-backup-error-topic"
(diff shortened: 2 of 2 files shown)


terraform/modules/lambda_s3/variables.tf
```diff
@@ -1,3 +1,9 @@
+# Create SNS topic subscription for each developer
+variable "developer_emails" {
+  type    = list(string)
+  default = ["anthony.hashemi@nationalarchives.gov.uk", "david.mckee@dxw.com"]
```
anthonyhashemi, line 4: followup PR will setup env vars to be passed in so we dont have store sensitive data such as email addresses in the repo, and then we will set up our other system email endpoints we want to send to as well.

jacksonj04: I'm happy for these to exist in the repo given that they're quasi-public anyway (we send our commits from them, so they're not secret they're just not widely published). And as you say, can tidy up in future PR so they're not needed in the codebase at all.

anthonyhashemi: yeah these arent the end of the world, and kept them here in this PR to keep things soewhat similar to before but I already have the followup PR ready based off this branch, which removes these and aexpects secrets from github. Will allow us to specify other email addfresses for different integrations which we might not want to be visible

dragon-dxw (approved):

PR #223, opened 2023-06-19
Fix function names being monitored for error alerts

anthonyhashemi:
Fix function names being monitored for error alerts

### DIFF
--- a/terraform/modules/lambda_s3/queue.tf
+++ b/terraform/modules/lambda_s3/queue.tf
@@ -365,7 +365,23 @@ resource "aws_lambda_event_source_mapping" "sqs_validated_xml_event_source_mappi
 
 
 locals {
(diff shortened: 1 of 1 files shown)


terraform/modules/lambda_s3/queue.tf
```diff
@@ -365,7 +365,23 @@ resource "aws_lambda_event_source_mapping" "sqs_validated_xml_event_source_mappi
...
+      "make-replacements",
+      "determine-replacements-abbreviations",
+      "determine-oblique-references",
+      "determine-legislation-provisions",
+      "xml-validate",
+      "push-enriched-xml",
+      "db-backup",
+      "update-rules-processor",
+      "update-legislation-table"
+    ] : "${local.name}-${local.environment}-${function_name}"
```
anthonyhashemi, line 383: we need the prefix, as otherwise we dont find the actual lambda function names on each environment!

dragon-dxw (approved):

PR #224, opened 2023-06-20
Bug/fix network issues and other deployment issues/1020

anthonyhashemi:
https://trello.com/c/76QZXuit/1020-fix-enrichment-networking

Changes:

This adds the security group definitions to fix the network and in doing so upgrades terraform, as well as some tweaks to fix certain lambda dependencies that were missing (wasnt a problem in production due to a deployment bug we have but that will be dealt with separately)

UPDATE: Verified that this works on staging by manually changing the ci file but have removed the change now and will merge


.github/workflows/ci_lint_and_test.yml
```diff
@@ -128,3 +128,14 @@ jobs:
       account_id: ${{ secrets.PROD_ACCOUNT_ID }}
       workspace: tna-prod
       app_env: production
+
+  deploy_staging:
```
anthonyhashemi, line 132: just for ease of testing - will remove before merging.

src/caselaw_extraction/requirements.txt
```diff
@@ -7,3 +7,4 @@ pandas==1.4.1
 psycopg2-binary==2.9.3
 beautifulsoup4==4.10.0
 lxml==4.9.2
+sqlalchemy==2.0.4
```
anthonyhashemi, line 10: multiple lambdas were missing this, now that they use the utils shared folder which depends on it.

Reason things hadnt been broken on prod must have been due to the slightly broken deployment process which doesnt always update the lambda to use the latest image - we have a ticket for that which we should get done ASAP!

We also have related ticket to sort out aws lambda shared layers

terraform/modules/network/vpc.tf
```diff
@@ -16,6 +16,24 @@ module "vpc" {
   create_database_nat_gateway_route  = true
   #create_database_internet_gateway_route = true
 
+  manage_default_security_group = true
+  default_security_group_egress = [
```
anthonyhashemi, line 20: main changes to allow changes out of the lambdas

terraform/modules/data/rds.tf
```diff
@@ -8,11 +8,12 @@ module "metadata-db" {
...
 
-  vpc_id                = var.vpc_id
-  subnets               = data.aws_subnets.database.ids
-  create_security_group = true
-  allowed_cidr_blocks   = [for s in data.aws_subnet.private : s.cidr_block]
-  deletion_protection   = local.db[local.environment].deletion_protection
+  vpc_id                  = var.vpc_id
+  subnets                 = data.aws_subnets.database.ids
+  create_security_group   = true
+  allowed_security_groups = [var.default_security_group_id]
```
anthonyhashemi, line 14: main change to allow inbound request to the rds database from the vpc the lambdas are in

terraform/main.tf
```diff
@@ -1,16 +1,16 @@
 
 terraform {
+  required_version = ">=1.4"
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = ">= 3.69.0, <= 4.4.0"
+      version = ">=5.3.0,<6.0.0"
```
anthonyhashemi, line 7: all the terraform related version upgrades are so we could utilise the new vpc aws module which the has better configurability for ingress, egress rules

dragon-dxw (approved):

rjw1 (approved):
The changes look fine. I do wonder if we should squash some of the commits. Also maybe the lamda fix should be seperate PR. 

We also at some point should discuss the use of the terraform lock file but thats out of scope of this PR.

anthonyhashemi:
> The changes look fine. I do wonder if we should squash some of the commits. Also maybe the lamda fix should be seperate PR.
> 
> We also at some point should discuss the use of the terraform lock file but thats out of scope of this PR.

Yeah I will squash commits, and agree the lambda fix can be separate. I just wanted to get a working pipeline on staging and I found those issues 1 by 1. Will remove non network stuff from this branch and put in a separate one 👍

anthonyhashemi:
> The changes look fine. I do wonder if we should squash some of the commits. Also maybe the lamda fix should be seperate PR.
> 
> We also at some point should discuss the use of the terraform lock file but thats out of scope of this PR.

And yes that would be good to discuss the terraform lock

PR #227, opened 2023-07-03
Force enrichment to identify namespace

dragon-dxw:
https://trello.com/c/lY169aWw/1079-enrichment-error-validation-for-ewfc-2023-9

A namespace can be defined in XML in either the node that uses it or any parent node.
If the `uk` namespace is defined in every node it is used in, rather than the root node,
then adding a new `uk:tna-enrichment-engine` node leaves that namespace undefined,
so we must define it ourselves.


dragon-dxw:
[Enrichment Error Validation for ewfc-2023-9](https://trello.com/c/lY169aWw/1079-enrichment-error-validation-for-ewfc-2023-9)

src/tests/lambda_tests/determine_legislation_provisions/test_determine_legislation_provisions.py
```diff
@@ -0,0 +1,34 @@
...
+    add_timestamp_and_engine_version,
+)
+
+sample_xml_1 = (
+    "<xml><proprietary/><FRBRManifestation><FRBRdate/></FRBRManifestation></xml>"
+)
+sample_xml_2 = '<xml xmlns:uk="https://caselaw.nationalarchives.gov.uk/akn"><proprietary/><FRBRManifestation><FRBRdate/></FRBRManifestation></xml>'
+
+
+def test_engine_version_valid_xml_without_namespace_is_valid():
```
anthonyhashemi, line 17: Could we add docstrings for these tests, as there is quite a bit of setup and the assertion is also non trivial?

src/tests/lambda_tests/determine_legislation_provisions/test_determine_legislation_provisions.py
```diff
@@ -0,0 +1,34 @@
...
+import re
+
+import lxml
+
+os.environ["DEST_BUCKET"] = "PLACEHOLDER"
+from lambdas.determine_legislation_provisions.index import (  # noqa: E402
+    add_timestamp_and_engine_version,
+)
+
+sample_xml_1 = (
```
anthonyhashemi, line 11: can we move these into the tests themselves since they are only used by one test each?

src/lambdas/determine_legislation_provisions/index.py
```diff
@@ -39,7 +39,10 @@ def add_timestamp_and_engine_version(file_data):
     enriched_date = soup.new_tag(
         'FRBRdate date="{}" name="tna-enriched"'.format(today_str)
     )
-    enrichment_version = soup.new_tag("uk:tna-enrichment-engine")
+    enrichment_version = soup.new_tag(
```
anthonyhashemi, line 42: I read the trello card and have a vague understanding on what caused the error and what led us to this change but could you summarise why only the document , `ewfc-2023-9`, was causing it and how this change gets rid of the bug down the chain in the validation function?

```
[ERROR] 2023-06-27T12:28:32.356Z ac84435c-76d8-4ffc-bae1-a7b6fc497a84 Exception: Namespace prefix uk on tna-enrichment-engine is not defined, line 46, column 26 (<string>, line 46)
```

anthonyhashemi (approved):
Approving with discussion from slack for visiblity if we ever come across anything around this again:

We're not quite sure why the particular judgment failed -- it had the header at one point the chain, and then it did not.

We think Beautiful Soup could have messed with where namespaces are stored or something along those lines?

Either way this deals with it and isnt worth chasing down this odd root cause.

PR #229, opened 2023-08-04
Bug/make replacements fails for self closing header element/1227

anthonyhashemi:
https://trello.com/c/gbMYrILW/1227-enrichment-enrichment-replacement-lambda-failing-for-legacy-tribunals


jacksonj04 (approved):
My innate distrust of manipulating XML by regex aside, this looks a solid replacement for what was there before.

PR #230, opened 2023-09-22
Fcl 151 enrichment broken for documents without header tags

anthonyhashemi:
**Bug details:**

Enrichment was failing to make replacements for for documents without header tags. Spcifically erroring with:
```
[ERROR] IndexError: list index out of range
Traceback (most recent call last):
  File "/var/task/index.py", line 219, in handler
    process_event(sqs_rec)
  File "/var/task/index.py", line 135, in process_event
    judgment_split[2], replacement_file_content
[ERROR] IndexError: list index out of range Traceback (most recent call last):   File "/var/task/index.py", line 219, in handler     process_event(sqs_rec)   File "/var/task/index.py", line 135, in process_event     judgment_split[2], replacement_file_content
```

**Changes in this PR:**

- This fixes this bug by fixing the splitting logic.
- Also refactored the main lambda function, extracting out the main logic away from the s3 code and removed unnecessary noisy print statements that were making the logs horrible to go through and debug.

Still need to:
- Write a test for the newly extracted logic function

### DIFF
--- a/src/lambdas/make_replacements/index.py
+++ b/src/lambdas/make_replacements/index.py
@@ -3,12 +3,10 @@
 import json
 import logging
 import os
-import re
 
 import boto3
-from bs4 import BeautifulSoup
 
-from replacer.make_replacments import split_text_by_closing_header_tag
+from replacer.make_replacments import make_post_header_replacements
 
 LOGGER = logging.getLogger()
 LOGGER.setLevel(logging.DEBUG)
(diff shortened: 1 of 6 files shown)


src/tests/replacer_tests/test_make_replacements.py
```diff
@@ -1,6 +1,30 @@
...
+        f"{FIXTURE_DIR}/ewhc-ch-2023-257_replacements.txt", "r", encoding="utf-8"
+    ).read()
+    expected_file_content = open(
+        f"{FIXTURE_DIR}/ewhc-ch-2023-257_enriched_stage_1.xml", "r", encoding="utf-8"
+    ).read()
+
+    content_with_replacements = make_post_header_replacements(
+        original_file_content, replacement_content
+    )
+    assert content_with_replacements == expected_file_content.strip()
```
anthonyhashemi, line 27: integration test for the business logic of this lambda

src/tests/replacer_tests/test_make_replacements.py
```diff
@@ -16,4 +40,4 @@ def test_split_text_by_closing_header_tag(self, closing_header_tag):
     @pytest.mark.parametrize("invalid_closing_header_tag", ["</foo>", "<foo/>"])
     def test_no_split(self, invalid_closing_header_tag):
         file_content = f"ABC{invalid_closing_header_tag}DEF"
-        assert split_text_by_closing_header_tag(file_content) == [file_content]
+        assert split_text_by_closing_header_tag(file_content) == ["", "", file_content]
```
anthonyhashemi, line 43: unit test for the bug fix

src/replacer/make_replacments.py
```diff
@@ -1,6 +1,139 @@
...
 import re
 from typing import List
 
+from bs4 import BeautifulSoup
+
+LOGGER = logging.getLogger()
+LOGGER.setLevel(logging.DEBUG)
+
+
+def make_post_header_replacements(
```
anthonyhashemi, line 12: all of this just moved over from the lambda file to keep the logic separate from aws infrastructure

src/lambdas/make_replacements/index.py
```diff
@@ -40,47 +38,6 @@ def upload_contents(source_key, text_content):
     object.put(Body=text_content)
 
 
-def detect_reference(text, etype):
```
anthonyhashemi, line 43: all moved to logic file

src/replacer/make_replacments.py
```diff
@@ -11,4 +11,4 @@ def split_text_by_closing_header_tag(file_content: str) -> List[str]:
     for pattern in header_patterns:
         if pattern in file_content:
```
dragon-dxw, line 145: It feels like a type error to be testing on "is string in other_string" and then using that string as a part of a regular expression. We should probably consider using `file_content.partition(pattern)`

Appreciate this isn't your code.

dragon-dxw (approved):

src/replacer/make_replacments.py
anthonyhashemi: good shout, have done so

PR #250, opened 2023-11-11
Run pre-commit in CI to check things, instead of firing them individually

jacksonj04:
This simplifies the test task by removing a bunch of independent tasks into a single "run pre-commit" job. This also improves parity between the checks run locally at commit-time and those running on CI.

### DIFF
--- a/.github/workflows/ci_lint_and_test.yml
+++ b/.github/workflows/ci_lint_and_test.yml
@@ -22,9 +22,29 @@ defaults:
     shell: bash
 
 jobs:
-  source_lint_and_test:
-    runs-on: ubuntu-latest
 
+  run-pre-commit:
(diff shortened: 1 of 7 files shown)


.github/workflows/ci_lint_and_test.yml
```diff
@@ -36,18 +47,6 @@ jobs:
...
-      - name: pip install formatting dependencies
-        run: |
-          pip3 install black==22.8.0 isort
-
-      - name: Python Black Formatting Check
-        id: run_python_black
-        run: black . --extend-exclude "^/get-pip.py" --check --diff
-
-      - name: Python isort Order Check
-        id: run_python_isort
```
dragon-dxw, line 48: what replaces isort?

PR #285, opened 2023-11-20
[ignore, was 3.12 again] Python 3.10 to see if that helps.

dragon-dxw:
Trying to see if moving to 3.10 helps.


dragon-dxw:
It wasn't even 3.10

PR #288, opened 2023-11-21
Renovate test requirements.txt

dragon-dxw:
Renovate skips some directories: see https://github.com/renovatebot/renovate/discussions/12755
No idea if this actually works.


jacksonj04 (approved):
I _think_ this is a working solution with no unwanted side-effects.

PR #308, opened 2023-11-21
Remove yum: no longer supported in modern Ubuntu

dragon-dxw:


dragon-dxw:
maybe this PR should be called "yuck" :P

jacksonj04 (approved):

PR #336, opened 2023-12-05
Update image-based Lambda functions at deploy time

### LINKED ISSUE #107: Lambda images are built and not always used
Docker images are built and stored in ECR. These are then used by lambda. 
Docker images are tagged with 'latest'. This scenario occurs:
* An image is built and stored in ECR, with the tag 'latest'
* Lambda has the configuration to use the image with a tag of 'latest' so it is using it. 

Later a new version of the image is deployed and is now called 'latest', the older image is there but untagged. Lambda is still using it because it doesn't track the change. Terraform would not update this because there are no infrastructure changes to deploy. 
The way around this would be to use versioned images without using latest as a tag.

jacksonj04:
Previously, although the Docker images behind these Lambdas were being updated and re-tagged, the Lambda functions themselves were not being told about the code updates.

They are now instructed to redeploy once images are updated.

Fixes #107.

### DIFF
--- a/.github/workflows/deploy_single_environment.yml
+++ b/.github/workflows/deploy_single_environment.yml
@@ -138,59 +138,68 @@ jobs:
           docker buildx build --load -t tna-s3-tna-ecr-repository-${{ secrets.app_env }} .
           docker tag tna-s3-tna-ecr-repository-${{ secrets.app_env }}:latest ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com/tna-s3-tna-ecr-repository-${{ secrets.app_env }}:latest
           docker push ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com/tna-s3-tna-ecr-repository-${{ secrets.app_env }}:latest
-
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy_single_environment.yml
```diff
@@ -138,59 +138,68 @@ jobs:
           docker buildx build --load -t tna-s3-tna-ecr-repository-${{ secrets.app_env }} .
           docker tag tna-s3-tna-ecr-repository-${{ secrets.app_env }}:latest ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com/tna-s3-tna-ecr-repository-${{ secrets.app_env }}:latest
           docker push ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com/tna-s3-tna-ecr-repository-${{ secrets.app_env }}:latest
-
+          aws lambda update-function-code --function-name tna-s3-tna-${{ secrets.app_env }}-determine-replacements-caselaw --image-uri ${{ secrets.account_id }}.dkr.ecr.eu-west-2.amazonaws.com/tna-s3-tna-ecr-repository-${{ secrets.app_env }}:latest
```
dragon-dxw, line 141: wierd this one doesn't have a repository suffix like the others, but I assume that's correct.

dragon-dxw (approved):

PR #339, opened 2023-12-06
Move back to semantic versioning

dragon-dxw:
Whilst there are advantages to having a single number for working out whether a version is new or not (cast to int, done) it is not that much more conceptually difficult to check the major part of the version number, and that does not violate the schema. Whilst it's likely no-one is using it, we probably shouldn't just change it.

- \[x\] Update CHANGELOG.md


anthonyhashemi (approved):
Makes sense to me

PR #344, opened 2023-12-11
Fix/bad test hash

dragon-dxw:
The reinsertion of ref tags had a bug where multiple replacements in a paragraph could be handled out-of-order, and the tag replacement code did not expect that.

- \[x\] Update CHANGELOG.md


src/replacer/second_stage_replacer.py
```diff
@@ -36,6 +36,11 @@ def replace_references(
     :param reference_replacements: list of dict of references and replacement information
     :return: string with replaced references
     """
+
+    # Splitting the text fails if the points are not sorted
+    reference_replacements = sorted(
```
dragon-dxw, line 41: This is the only substantial line -- we sort the reference_replacements before we use them.

jacksonj04 (approved):

PR #349, opened 2023-12-13
Better XML generation -- 

dragon-dxw:
Handle the XML replacements as actual XML (`create_tag` and `create_tag_string`), and escape some input strings that aren't from the XML.

Not sure if we should just do the string escaping separatly and maybe not do the create_tag work?

- \[x\] Update CHANGELOG.md

### DIFF
--- a/src/oblique_references/oblique_references.py
+++ b/src/oblique_references/oblique_references.py
@@ -20,8 +20,9 @@ The pipeline returns a dictionary containing the detected oblique reference, its
 """
 
 import re
-from typing import Any, Dict, List, Tuple, Union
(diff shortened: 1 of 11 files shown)


src/oblique_references/oblique_references.py
```diff
@@ -146,6 +149,16 @@ def match_act(
     return matched_act
 
 
+def make_valid_tag(
```
jacksonj04, line 152: This entire function is unused and can go away; it was one of our early stabs which didn't handle inline tags.

jacksonj04 (approved):
One bit of dead code to nuke, otherwise LGTM

PR #353, opened 2023-12-14
Fix case where `replace_references_by_paragraph` was treating XML as HTML

jacksonj04:
When you initialise BeautifulSoup, you can provide a parser argument. This defaults to HTML, but the rest of Enrichment provides an argument of `"xml"`, which uses Python's default XML parser. In this case, however, the argument was "lxml", which sounds like it should use `lxml.etree`, but it doesn't. Instead it uses `lxml.html`, which converts all tags to lowercase. Unfortunately, XML's tags are case sensitive and this causes explosions with some of our case-sensitive tags such as `authorialNote`.

However, when we remove the `l`, we hit another problem. BeautifulSoup is a forgiving parser, sees that there are no namespaces defined and just throws them away. We wrap the tag in another tag with a defined `uk:` namespace, which BeautifulSoup then pulls just the paragraph from.

### DIFF
--- a/src/replacer/second_stage_replacer.py
+++ b/src/replacer/second_stage_replacer.py
@@ -7,7 +7,7 @@ Handles the replacements of oblique references and legislation provisions.
 
 import re
 from itertools import groupby
-from typing import Dict, List, Tuple, Union
+from typing import Dict, Iterable, List, Tuple, Union
 
 from bs4 import BeautifulSoup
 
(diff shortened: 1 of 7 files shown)


dragon-dxw (approved):

src/replacer/second_stage_replacer.py
```diff
@@ -64,6 +64,14 @@ def replace_references(
     return enriched_text
 
 
+def create_replacement_paragraph(paragraph_string, paragraph_reference_replacements):
```
jacksonj04, line 67: Can we add some type annotations to make it clearer exactly what we're expecting and to let mypy shout at us if we get it wrong?

src/utils/tests/compare_xml.py
```diff
@@ -0,0 +1,29 @@
...
+
+    if isinstance(b, str):
+        b = b.encode("utf-8")
+    canon_b = canonical_xml(b)
+    if canon_a != canon_b:
+        for i, (char_a, char_b) in enumerate(zip(canon_a, canon_b)):
+            if char_a != char_b:
+                break
+        raise AssertionError(
+            f"xml mismatch at {i}\nbefore: {canon_a[i-width:i]}\n first: {canon_a[i:i+width]}\nsecond: {canon_b[i:i+width]}"
```
jacksonj04, line 30: Nice.

PR #355, opened 2023-12-15
See process of a judgment through the enrichment pipeline

dragon-dxw:
Type in URL (in either slash or hyphen form) and link directly to AWS page for XML or cloudwatch logs. 
image


jacksonj04 (approved):
LGTM, my only comment would be if it's worth maybe having a `/tools` directory or similar for things like this instead of cluttering the root?

PR #363, opened 2023-12-20
Add thorough type annotations

dragon-dxw:
Enrichment has a lot of moving parts which need to be able to reliably pass data between them. Adding thorough type annotations gives us massively increased confidence in what is actually happening, reducing scope for errors to creep in. Amongst other things, this PR:

- Adds a bunch of type extensions to mypy for greater introspection ability
- Improves module discovery so static analysis can get a better handle on imports
- Makes use of [AWS Lambda Powertools](https://docs.powertools.aws.dev/lambda/python/latest/) to give us more strongly typed interactions with Lambda trigger events
- Creates some new type classes so we can be more confident in what we're passing around
- Deduplicates some functions where obvious to assist in typing
- Makes a bunch of type declarations

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -61,6 +61,7 @@ repos:
             types-lxml,
             types-beautifulsoup4,
             "boto3-stubs[essential, secretsmanager]",
+            aws_lambda_powertools,
           ]
         exclude: tests/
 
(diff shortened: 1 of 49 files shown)


dragon-dxw:
Reviewed up to WIP after "Add Typing" with one comment -- version number regression.

jacksonj04 (approved):

PR #382, opened 2024-01-03
Swap to using Poetry for dependency management in fetch_xml

jacksonj04:
This is a **spike** on if we can use Poetry to better manage dependencies for lambda functions, by vendoring dependencies at deploy time prior to bundling.


jacksonj04:
Closing this spike as this is going to need a lot of busywork which feels quite hacky for not much tangible benefit. It would make things more mechanical, but I'm not convinced it's the right approach.

PR #384, opened 2024-01-04
Don't import spacy if we only use it for typing

dragon-dxw:


jacksonj04 (approved):
This means we can also roll back adding spacy to all the requirements!

(Ngl, I'd totally forgotten you could do this)

PR #385, opened 2024-01-04
Revert "Add spacy to all the requirements"

dragon-dxw:
This reverts commit 5e30979763fb04132b0cbd9e262351fd1d9ea1dd.

We don't need it because we're only importing spacy for type checking.

- \[ \] Increase the version number in src/lambdas/determine_legislation_provisions/index.py
- \[ \] Update CHANGELOG.md

### DIFF
--- a/src/database/requirements.txt
+++ b/src/database/requirements.txt
@@ -1,4 +1,3 @@
 pandas==2.1.4
 psycopg2==2.9.9
 sqlalchemy==2.0.24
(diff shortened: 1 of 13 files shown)


dragon-dxw:
Ready to review but going to leave merging it a few hours at least to see if our `if TYPE_CHECKING:` fix calms the situation. This isn't urgent.

PR #388, opened 2024-01-05
Only import mypy boto3 if typechecking

dragon-dxw:


jacksonj04 (approved):
That ..database is weird, otherwise :shipit:

PR #396, opened 2024-01-08
Remove all NLP typing

dragon-dxw:
We're getting `module 'spacy.lang' has no attribute 'en'` errors in determine-replacements-legislation. There's probably a reasonable way of doing this but let's not right now.


dragon-dxw:
It appears to be working?

jacksonj04 (approved):

PR #401, opened 2024-01-10
Auto unlock if timeout

dragon-dxw:
We set an `expires` value that isn't 0; that will make it any lock expire at midnight. It's set to 3600 to make it less painful if/when we change the behaviour to expiring in a number of seconds.

This will need updating in api client and priv api.


jacksonj04 (approved):
LGTM. If enrichment doesn't complete within an hour we should absolutely unlock, given that we say ~6 minutes is the estimated runtime.

PR #431, opened 2024-02-08
Request lock release whilst sending XML to marklogic

dragon-dxw:
Benefits from the updated version of the privileged API to ensure that the document is always unlocked after use -- will break unlocking if that other one hasn't been merged.

https://github.com/nationalarchives/ds-caselaw-privileged-api/pulls

### DIFF
--- a/src/lambdas/push_enriched_xml/index.py
+++ b/src/lambdas/push_enriched_xml/index.py
@@ -60,6 +60,7 @@ def patch_judgment_request(
         f"{api_endpoint}judgment/{query}",
         auth=HTTPBasicAuth(username, pw),
         data=data.encode(),
+        params={"unlock": True},
     )
     print(response)
     response.raise_for_status()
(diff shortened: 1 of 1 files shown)


src/lambdas/push_enriched_xml/index.py
```diff
@@ -110,9 +111,6 @@ def process_event(sqs_rec: SQSRecord) -> None:
         api_endpoint, judgment_uri, file_content, API_USERNAME, API_PASSWORD
     )
 
-    # release the lock
-    release_lock(api_endpoint, judgment_uri, API_USERNAME, API_PASSWORD)
-
```
dragon-dxw, line 115: This is no longer needed because the lock will be released as a part of patching the file.

CristinaRO (approved):
Good work! We'll let you do the merging in the needed order, but the other PR is also approved https://github.com/nationalarchives/ds-caselaw-privileged-api/pull/303.

PR #434, opened 2024-02-14
Add UKAIT rules to CSV

dragon-dxw:
This will need deploying by copying the CSV file to S3 manually.

We first copy the rules that are actually in production to the repo, then add a line for UKAIT. We then write a test for the CSV file format to make sure all lines are picked up.

Additional refactoring exists in https://github.com/nationalarchives/ds-caselaw-data-enrichment-service/pull/433


pezholio (approved):
LGTM - weird that the Coverage checker is failing, but I guess that's expected

PR #436, opened 2024-04-23
Terraform housekeeping

Stretch96:
* Adds a terraform-docs DotFile
* Removes what appears to be old commented code
* Collates `locals` into single files
* Pins module versions to ensure resources aren't unintentionally modified
* Makes use of `jsonencode` rather than JSON strings for easier readability
* Ran a `terraform fmt` to format the HCL


terraform/main.tf
```diff
@@ -13,15 +12,15 @@ terraform {
...
     key    = "ds-infrastructure-enrichment-pipeline/backend.tfstate"
     region = "eu-west-2"
   }
 }
 
 module "lambda_s3" {
   source = "./modules/lambda_s3"
-  name   = "tna"
+
+  name = "tna"
```
dragon-dxw, line 23: There's a part of me that doesn't like the name 'tna' as not being descriptive, but it's an old name that's not worth changing.

terraform/modules/data/locals.tf
```diff
@@ -7,4 +11,9 @@ locals {
       deletion_protection = true
     }
   }
+
+  tags = {
+    Environment = var.environment
+    Project     = "TNA judgement enrichment"
```
dragon-dxw, line 17: As a super-nitpicky comment: we use "judgment" (no central e) throughout whenever referring to a legal document as a noun.

dragon-dxw: I'd also generally refer to the project as "Find Case Law" so it's distinct from "the Transformation Engine" etc.

terraform/modules/lambda_s3/queue.tf
```diff
@@ -65,7 +60,10 @@ resource "aws_sqs_queue" "replacements-queue" {
   message_retention_seconds  = 86400
   receive_wait_time_seconds  = 10
   sqs_managed_sse_enabled    = true
-  redrive_policy             = "{\"deadLetterTargetArn\":\"${aws_sqs_queue.replacements_dlq_queue.arn}\",\"maxReceiveCount\":4}"
+  redrive_policy = jsonencode({
```
dragon-dxw, line 63: <3

terraform/modules/network/local.tf
```diff
@@ -1,18 +0,0 @@
-locals {
```
dragon-dxw, line 1: Curious as to why this isn't needed.

terraform/main.tf
Stretch96: Yep if we change that, it'll delete/recreate a lot of the resources

terraform/modules/data/locals.tf
Stretch96: I'll need to check that this tag isn't being used as a filter (eg. for billing) before we can change this - I'll add a Trello card 👍

PR #446, opened 2024-04-26
Fix Jump Host SSM DHMC policy

Stretch96:
* The resorce for the ssm settings should be the ARN of the dhmc setting


rjw1 (approved):

CristinaRO (approved):
Chatted in Slack.

PR #496, opened 2024-07-03
[FCL-176] Tooling configuration audit 

jacksonj04:
A series of commits which together make all of FCL's repositories behave more consistently with regard to tooling.

## Jira

FCL-176

## What?

- [x] `.pre-commit-config.yaml` does not exclude directories which don't exist, doesn't set default hooks where they aren't needed, and doesn't specify auto-update behaviour
- [x] The `pre-commit-hooks` repo includes the new, broader set of recommended checks
- [x] CI excludes the pre-commit branch check
- [x] Python formatting is done using [ruff](https://docs.astral.sh/ruff/)
- [x] `detect-secrets` is removed (deprecated in favour of GitHub's more comprehensive approach)
- [x] prettier config has been harmonised


setup.cfg
```diff
@@ -1,26 +1,3 @@
-[flake8]
```
dragon-dxw, line 1: Did ruff format delete these?

jacksonj04: I did, fixing which commit they end up in!

dragon-dxw (approved):
All looks the best kind of boring. ✅

PR #513, opened 2024-07-08
[FCL-176] Add Ruff linting

jacksonj04:
Following on from #496, add ruff's linting (as opposed to just formatting) and work through the necessary tweaks to make it happy.


src/lambdas/db_backup/index.py
```diff
@@ -14,7 +14,7 @@ def lambda_handler(event: EventBridgeEvent, context: LambdaContext) -> None:
     # Create RDS and S3 clients
     rds = boto3.client("rds")
     db = event["db-name"]
-    now = datetime.now()
+    now = datetime.now(tz=UTC)
```
dragon-dxw, line 17: Depending on when snapshots are taken (if within one hour of midnight), this might lead to the overwriting of a days snapshot around one of the daylight savings.

dragon-dxw, line 17: Fixes a potential issue where backups could overwrite each other if saved within one hour of midnight around daylight saving.

src/lambdas/determine_legislation_provisions/index.py
```diff
@@ -44,7 +44,7 @@ def add_timestamp_and_engine_version(
     Add today's timestamp and version at time of enrichment
     """
     soup = BeautifulSoup(file_data, "xml")
-    today = datetime.datetime.now()
+    today = datetime.datetime.now(tz=datetime.UTC)
     today_str = today.strftime("%Y-%m-%dT%H:%M:%S")
     enriched_date = soup.new_tag(f'FRBRdate date="{today_str}" name="tna-enriched"')
```
dragon-dxw, line 49: Bringing the timestamp to UTC is definately valid.
I'd like to see a timezone on the strftime but that might break compatibility, so sensible not to right now.

src/lambdas/update_legislation_table/update_legislation_table/tests/test_fetch_legislation.py
```diff
@@ -26,7 +26,7 @@ def set_env_vars():
     del os.environ["SPARQL_PASSWORD"]
 
 
-frozen_date = datetime.datetime(2023, 4, 13)
+frozen_date = datetime.datetime(2023, 4, 13, tzinfo=datetime.UTC)
```
dragon-dxw, line 29: On L41, mock_datetime.today is set using this value, but we no longer use datetime.today(). This test is no longer frozen.

src/database/db_connection.py
```diff
@@ -1,12 +1,21 @@
...
 
-from typing import Any
+from typing import Any, NamedTuple
 
 import pandas as pd
 import psycopg2
 from sqlalchemy import Connection
 
 
+class MatchedRule(NamedTuple):
```
dragon-dxw, line 10: This feels quite a bold change. Good though.

dragon-dxw (approved):

PR #542, opened 2024-09-19
Make Legislation integration test run against a canned version

dragon-dxw:
There was an assertion failing in `test_fetch_legislation.py::test_fetch_legislation_integration`. This connects to the Legislation SPARQL endpoint. It looks like the order of the rows returned has changed, and the URL returned is http://www.legislation.gov.uk/ukpga/2023/10 rather than http://www.legislation.gov.uk/ukpga/2023/10/enacted.

It's possible that this change of structure will impact the rest of enrichment, but it's not clear that it will. It's possible it's related to the issues we've been encountering with URLs recently. We've canned the response from legislation.

- \[x\] Update CHANGELOG.md

https://national-archives.atlassian.net/browse/FCL-288

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## v6.0.1 (2024-07-16)
 
+- Fix CI by reflecting changed output from legislation SPARQL output
+
 ### Refactor
 
 - **FCL-176**: explicitly set timezones to UTC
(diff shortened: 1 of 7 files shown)


jacksonj04 (approved):
Beautiful. Nice tidying up of test invocation as well!

PR #670, opened 2024-11-25
[FCL-490] Use XML parsing for first stage replacements

dragon-dxw:

- [ ] Update CHANGELOG.md


src/replacer/replacer_pipeline.py
```diff
@@ -6,9 +6,33 @@
...
 
-from utils.proper_xml import create_tag_string
+from utils.proper_xml import create_tag_string, replace_string_with_tag
 from utils.types import Replacement, XMLFragmentAsString
 
+JUNK_REGEX = r"</judgment>\s*</akomaNtoso>\s*$"
+BAD = '="<'
+
+
+def assert_not_bad(s):
```
dragon-dxw, line 16: We could replace this with an lxml scan. We might not need it every time.

src/replacer/replacer_pipeline.py
```diff
@@ -40,9 +64,13 @@ def replacer_caselaw(file_data: XMLFragmentAsString, replacement: Replacement) -
     if year:
         attribs["uk:year"] = year
     attribs["uk:origin"] = "TNA"
-    replacement_string = create_tag_string("ref", html.escape(replacement[0]), attribs)
 
-    return XMLFragmentAsString(str(file_data).replace(replacement[0], replacement_string))
+    ### DRAGON: use utils.proper_xml.create_tag_string when creating. But for now, focus on making it not change attributes, only XML running text
```
dragon-dxw, line 68: TODO

src/tests/fixtures/broken.txt
```diff
@@ -0,0 +1,147 @@
...
+{"case": ["[1980] 1 WLR 277", "[1980] 1 WLR 277", "1980", "#", false]}
+{"case": ["[1982] 1 Lloyd\u2019s Rep 570", "[1982] 1 Lloyd's Rep 570", "1982", "#", false]}
+{"case": ["[2007] 2 AC 353", "[2007] 2 AC 353", "2007", "#", false]}
+{"case": ["[1980] 1 Lloyd\u2019s Rep 75", "[1980] 1 Lloyd's Rep 75", "1980", "#", false]}
+{"case": ["[2011] 2 Lloyd\u2019s Rep 360", "[2011] 2 Lloyd's Rep 360", "2011", "#", false]}
+{"case": ["[2011] 2 Lloyd\u2019s Rep 370", "[2011] 2 Lloyd's Rep 370", "2011", "#", false]}
+{"case": ["[1995] 1 Lloyd\u2019s Rep 142", "[1995] 1 Lloyd's Rep 142", "1995", "#", false]}
+{"leg": ["Sale of Goods Act 1979", "http://www.legislation.gov.uk/id/ukpga/1979/54", "1979 c. 54"]}
+{"leg": ["Sale of Goods Act 1979", "http://www.legislation.gov.uk/id/ukpga/1979/54", "1979 c. 54"]}
+{"leg": ["Sale of Goods Act 1979", "http://www.legislation.gov.uk/id/ukpga/1979/54", "1979 c. 54"]}
```
dragon-dxw, line 147: Note: this is real data, and we should be cautious about using it.

src/tests/replacer_tests/test_make_replacements.py
```diff
@@ -13,6 +13,12 @@
 
 
 class TestMakePostHeaderReplacements:
+    def test_make_failing_post_header_replacements(self):
+        original_file_content = open(f"{FIXTURE_DIR}/broken.xml", encoding="utf-8").read()
```
dragon-dxw, line 17: consider `FIXTURE_DIR / xxx`

dragon-dxw: consider more descriptive fixture name

src/replacer/replacer_pipeline.py
```diff
@@ -87,13 +120,19 @@ def replacer_pipeline(
     :param REPLACEMENTS_ABBR: list of unique tuples of citation match and corrected citation
     :return: enriched XML file data
     """
+
+    assert_not_bad(file_data)
```
dragon-dxw, line 124: Just the once should do when not actively debugging.

src/utils/proper_xml.py
```diff
@@ -1,10 +1,54 @@
...
+    </xsl:template>
+
+    <!-- Copy everything else as-is -->
+    <xsl:template match="@*|node()">
+        <xsl:copy>
+            <xsl:apply-templates select="@*|node()"/>
+        </xsl:copy>
+    </xsl:template>
+</xsl:stylesheet>
+"""
```
dragon-dxw, line 36: Consider making this take multiple replacements at once for speed?

src/utils/tests/test_proper_xml.py
```diff
@@ -1,7 +1,65 @@
...
+                "<p><span>[2024] UKSC 1</span><span>[2024] UKSC 1</span></p>",
+                "[2024] UKSC 1",
+                "<cite/>",
+            ),
+            b"<p><span><cite/></span><span><cite/></span></p>",
+        )
+
+    @pytest.mark.xfail
+    def test_as_text_multi_in_tag(self):
+        """Ideally, this should have <cite/> - <cite/> in the output, instead of just doing the first; we could rerun until it runs clean"""
```
dragon-dxw, line 48: TODO

src/utils/tests/test_proper_xml.py
```diff
@@ -1,7 +1,65 @@
...
+
+class TestReplaceStringWithTag:
+    def test_as_text(self):
+        assert_equal_xml(
+            replace_string_with_tag("<p>In [2024] UKSC 1 ...</p>", "[2024] UKSC 1", "<ref blah='blah'>blah</ref>"),
+            b'<p>In <ref blah="blah">blah</ref> ...</p>',
+        )
+
+    @pytest.mark.xfail
+    def test_as_text_quotes(self):
```
dragon-dxw, line 15: Any insights as to the right way to do this would be appreciated.

dragon-dxw:
We verified that the issue is not reparsing already parsed things -- the failing judgment contains no <ref> tags nor any signatures of being enriched (versions, etc.)

Instead we now think that the enrichment process applies these two rules:

```
('[1971] 1 Q.B. 488', '[1971] 1 QB 488', '1971', '#', False)
('[1971] 1 QB 488', '[1971] 1 QB 488', '1971', '#', False)
```

And sees `[1971] 1 Q.B. 488`, converts that to `<ref id="[1971] 1 QB 488">...</ref>` and then on the second line to `<ref id="<ref id="...">...</ref></ref>`. Which is not valid.

src/lambdas/extract_judgement_contents/index.py
```diff
@@ -34,7 +34,7 @@ def process_event(sqs_rec: S3EventRecord):
     upload_contents(source_key, text_content)
 
 
-def extract_text_content(file_content: DocumentAsXMLString) -> str:
+def extract_text_content(file_content: DocumentAsXMLString) -> DocumentAsXMLString:
```
jacksonj04, line 37: Is this actually a string representation of XML? I think it's just a string of all the words in the XML.

src/utils/helper.py
```diff
@@ -9,7 +9,7 @@
 from utils.types import DocumentAsXMLString
 
 
-def parse_file(file_data: DocumentAsXMLString) -> str:
+def parse_file(file_data: DocumentAsXMLString) -> DocumentAsXMLString:
```
jacksonj04, line 12: Another instance where I think this is a string of text, not a string of XML.

src/oblique_references/oblique_references.py
```diff
@@ -241,11 +244,11 @@ def get_oblique_reference_replacements_by_paragraph(
     """
     soup = BeautifulSoup(file_content, "xml")
     paragraphs = soup.find_all("p")
-    all_replacements: list[dict] = []
+    all_replacements: list[LegislationReferenceReplacement] = []
```
jacksonj04, line 247: Should this be `LegislationReferenceReplacements`? instead of the `list[]`?

src/oblique_references/oblique_references.py
```diff
@@ -241,11 +244,11 @@ def get_oblique_reference_replacements_by_paragraph(
...
     all_legislation_dicts = []
 
     for paragraph_number, paragraph in enumerate(paragraphs):
-        replacements: list[dict] = []
+        replacements: list[LegislationReferenceReplacement] = []
```
jacksonj04, line 251: Should this be `LegislationReferenceReplacements`? instead of the `list[]`?

jacksonj04 (changes requested):
Couple of tiny tweaks, but this looks solid 👍🏻

src/tests/replacer_tests/test_make_replacements.py
```diff
@@ -13,6 +14,12 @@
 
 
 class TestMakePostHeaderReplacements:
+    def test_make_failing_post_header_replacements(self):
```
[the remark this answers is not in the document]
dragon-dxw: Write a small actual test that fails, should be OK.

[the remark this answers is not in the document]
dragon-dxw: Assert correct exception would be great.

src/lambdas/extract_judgement_contents/index.py
dragon-dxw: 👍

src/utils/helper.py
dragon-dxw: :+1:

src/oblique_references/oblique_references.py
dragon-dxw: use list[LegislationReferenceReplacement] throughout instead.

src/oblique_references/oblique_references.py
dragon-dxw: see above, invert.

src/tests/fixtures/broken.txt
jacksonj04: I'm not worried about this; nothing here is sensitive or secret, and it makes testing more reliable.

PR #803, opened 2025-02-13
Move types out of __init__.py file in types subfolder into a custom_types.py file

anthonyhashemi:
## Changes in this PR:

Move types out of `__init__.py` file in a `types` subfolder into a `custom_types.py` file for reasons:
- python stopped being able to import from the `/types/__init__.py` file for some reason last week
- no reason these should be defined in a init.py and not just a types file
- types is an internal python module and so types.py would be renaming an internal python module which is not possible ansd causes and issue (as seen in the precommit checks 
Screenshot 2025-02-13 at 15 05 50
for example in this pr to updated ruff precommit: https://github.com/nationalarchives/ds-caselaw-data-enrichment-service/actions/runs/13310142329/job/37170355518?pr=797

https://national-archives.atlassian.net/browse/FCL-672


dragon-dxw (approved):

dragon-dxw:
Pondering whether there should be a standardised naming structure for "here are the types" to avoid types shadowing.

PR #983, opened 2025-05-19
Fix replace_string_with_tag for when the xml fragment string includes uk namespace attributes

anthonyhashemi:
# Changes in this PR:

Fix replace_string_with_tag for when the xml fragment string includes uk namespace attributes

Don't love my current fix but I think the whole thing needs a rethink, but this should suffice for now

https://national-archives.atlassian.net/browse/FDD-35

- [x] Update CHANGELOG.md

### DIFF
--- a/src/tests/replacer_tests/test_replacer_pipeline.py
+++ b/src/tests/replacer_tests/test_replacer_pipeline.py
@@ -17,27 +17,27 @@ class TestCitationReplacer(unittest.TestCase):
         citation_match = "[2025] 1 All E.R. 123"  # incorrect citation
         corrected_citation = "[2025] 1 All ER 123"  # in practice, returned via the citation matcher
         year = "2025"
-        URI = "#"
+        uri = "#"
         is_neutral = "true"
-        text = "<z>In the judgment the incorrect citation is [2025] 1 All E.R. 123.</z>"
-        replacement_entry = (citation_match, corrected_citation, year, URI, is_neutral)
-        replaced_entry = replacer_caselaw(text, replacement_entry)
-        assert corrected_citation in replaced_entry
-        replacement_string = f'<ref xmlns:uk="https://caselaw.nationalarchives.gov.uk/akn" xmlns="http://docs.oasis-open.org/legaldocml/ns/akn/3.0" uk:type="case" href="{URI}" uk:isNeutral="{is_neutral}" uk:canonical="{corrected_citation}" uk:year="{year}" uk:origin="TNA">{citation_match}</ref>'
-        assert replacement_string in replaced_entry
+        text = "<z>In the judgment the incorrect citation is [2025] 1 All E.R. 123. <table uk:widths='1.38in 1.05in 0.91in 0.98in 0.98in 0.95in'></table></z>"
+        replacement_entry = (citation_match, corrected_citation, year, uri, is_neutral)
+        assert (
+            replacer_caselaw(text, replacement_entry)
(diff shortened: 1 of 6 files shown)


src/utils/proper_xml.py
```diff
@@ -41,13 +41,21 @@ def replace_string_with_tag(xml: XMLFragmentAsString, string: str, tag: str) ->
     Replace citations in running text, but not in XML attributes
     This is probably quite inefficient, being called once per replacement! But we can make it better later
     """
+    root_start = '<root xmlns:uk="https://caselaw.nationalarchives.gov.uk/akn">'
```
anthonyhashemi, line 44: the fix: wrap the xml fragment string with a root element that defines the uk namespace as otherwise lxml errors when we try to build an etree from string if the fragment contains uk namespace attributes

src/tests/replacer_tests/test_replacer_pipeline.py
```diff
@@ -17,26 +17,27 @@ def test_citation_replacer_1(self):
         citation_match = "[2025] 1 All E.R. 123"  # incorrect citation
         corrected_citation = "[2025] 1 All ER 123"  # in practice, returned via the citation matcher
         year = "2025"
-        URI = "#"
+        uri = "#"
         is_neutral = "true"
-        text = "<z>In the judgment the incorrect citation is [2025] 1 All E.R. 123.</z>"
-        replacement_entry = (citation_match, corrected_citation, year, URI, is_neutral)
+        text = "<z>In the judgment the incorrect citation is [2025] 1 All E.R. 123. <table uk:widths='1.38in 1.05in 0.91in 0.98in 0.98in 0.95in'></table></z>"
```
anthonyhashemi, line 22: to be honest this test change isn';t really needed now I have added the test on the smaller unit but think it is fine to keep

src/tests/replacer_tests/test_replacer_pipeline.py
```diff
@@ -17,26 +17,27 @@ def test_citation_replacer_1(self):
...
+            replaced_entry
+            == '<z xmlns:uk="https://caselaw.nationalarchives.gov.uk/akn">In the judgment the incorrect citation is <ref xmlns="http://docs.oasis-open.org/legaldocml/ns/akn/3.0" uk:type="case" href="#" uk:isNeutral="true" uk:canonical="[2025] 1 All ER 123" uk:year="2025" uk:origin="TNA">[2025] 1 All E.R. 123</ref>. <table uk:widths="1.38in 1.05in 0.91in 0.98in 0.98in 0.95in"/></z>'
+        )
 
     def test_citation_replacer_2(self):
         citation_match = "[2022] UKET 789123_2012"
         corrected_citation = "[2022] UKET 789123/2012"
         year = "2022"
-        URI = "#"
+        uri = "#"
```
anthonyhashemi, line 34: I also renamed URI to uri, as capitalised vars reserved for constants

src/utils/proper_xml.py
```diff
@@ -41,13 +41,28 @@ def replace_string_with_tag(xml: XMLFragmentAsString, string: str, tag: str) ->
     Replace citations in running text, but not in XML attributes
     This is probably quite inefficient, being called once per replacement! But we can make it better later
     """
-    root = lxml.etree.fromstring(xml)
+    uk_ns_1 = 'xmlns:uk="https://caselaw.nationalarchives.gov.uk/akn"'
+    uk_ns_2 = "xmlns:uk='https://caselaw.nationalarchives.gov.uk/akn'"
+    needs_wrapper = uk_ns_1 not in str(xml) and uk_ns_2 not in str(xml)
+    root_start = f"<root {uk_ns_1}>" if needs_wrapper else "<root>"
+    root_end = "</root>"
+    xml_wrapped = XMLFragmentAsString(root_start + str(xml) + root_end)
```
anthonyhashemi, line 49: the fix: wrap the xml fragment string with a root element that defines the uk namespace as otherwise lxml errors when we try to build an etree from string if the fragment contains uk namespace attributes.

However, only wrap it when the outer element doesnt already define the namespace, and then remove the outer root wrapper at the end of the function

jacksonj04 (approved):

PR #1059, opened 2025-06-27
Fix priv api uri issue

anthonyhashemi:
## Changes in this PR:

- [Fix mypy issue](https://github.com/nationalarchives/ds-caselaw-data-enrichment-service/commit/86a4aa9f84a88dbe37e236274dec5c26aa0352ba)

- [Remove redundant filename variable](https://github.com/nationalarchives/ds-caselaw-data-enrichment-service/commit/1a05b7afc924dac047c942256f96a2c1320ab4e7)

- [Stop replacing hyphens with slashes when storing the xml in s3 as causes issue for documents with hyphens in uri when trying to reverse the process when pushing back to judgment database](https://github.com/nationalarchives/ds-caselaw-data-enrichment-service/commit/68f61f9c3bbd4ad5baef032257b2595b76e24979) 

- [x] Update CHANGELOG.md

### DIFF
--- a/src/legislation_provisions_extraction/legislation_provisions.py
+++ b/src/legislation_provisions_extraction/legislation_provisions.py
@@ -63,7 +63,7 @@ def find_closest_legislation(legislations, sections, thr=30):
 
     # returns sections that are within a threshold distance from legs
     idx = np.argwhere(dist < thr)
-    section_to_leg = [(sections[i][1], legislations[j][1], sections[i][0][0]) for i, j in idx]  # type: ignore[has-type, misc]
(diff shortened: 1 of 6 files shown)


src/legislation_provisions_extraction/legislation_provisions.py
```diff
@@ -247,38 +249,6 @@ def provision_resolver(section_dict, matches, para_number):
     return resolved_refs
 
 
-def main(enriched_judgment_file_path, filename):
```
anthonyhashemi, line 250: removed as was unused (stumbled across it when trying to fix mypy precommit issue

dragon-dxw (approved):

PR #1070, opened 2025-07-01
Fully canonicalise final stage enriched xml before writing back to marklogic and bump to version 7.2.4

anthonyhashemi:
## Changes in this PR:

Fully canonicalise final stage enriched xml before writing back to marklogic and bump to version 7.2.4

- [x] Update CHANGELOG.md

### DIFF
--- a/src/lambdas/push_enriched_xml/index.py
+++ b/src/lambdas/push_enriched_xml/index.py
@@ -98,7 +98,9 @@ def process_event(sqs_rec: SQSRecord) -> None:
     file_content = DocumentAsXMLString(
         s3_client.get_object(Bucket=source_bucket, Key=source_key)["Body"].read().decode("utf-8"),
     )
-    canonical_xml = etree.canonicalize(file_content)
+    # Canonicalize using lxml.etree.tostring with method="c14n"
+    tree = etree.fromstring(file_content)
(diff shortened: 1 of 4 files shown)


src/lambdas/push_enriched_xml/index.py
```diff
@@ -98,7 +98,15 @@ def process_event(sqs_rec: SQSRecord) -> None:
     file_content = DocumentAsXMLString(
         s3_client.get_object(Bucket=source_bucket, Key=source_key)["Body"].read().decode("utf-8"),
     )
-    canonical_xml = etree.canonicalize(file_content)
+
+    # Canonicalize the xml with C14N 1.0 so that only the apex node defines the namespaces
+    # and the attributes are in the same order.
+    # This is important for the API to accept the XML.
+    # Note C14N 2.0 is not supported by the API and further the XML should not have
```
dragon-dxw, line 105: I find it utterly befuddling that C14N 2.0 can not be valid XML according to MarkLogic -- my understanding was that it's supposed to be a way of choosing between perfectly valid things which don't change the meaning of the XML in any way, so the structure at the end should be valid.

Super wierd.

dragon-dxw (approved):

PR #1109, opened 2025-08-04
Swap in codecov to replace codeclimate

jacksonj04:
To unclog tests, extract the replacement of Codeclimate with CodeCov from #1108.


anthonyhashemi:
Already have a PR for this @jacksonj04: https://github.com/nationalarchives/ds-caselaw-data-enrichment-service/pull/1108

PR #1955, opened 2026-06-11
Rewrite enrichment lambda pipeline into one lambda

anthonyhashemi:

## Changes in this PR:

- Consolidated enrichment lambda pipeline into one lambda
- Restructured certain functions/files into different locations
- Added a currently commented-out regression test with an example xml file and its enriched counterpart from before this change to confirm that we don't have a regression. Need to talk about this file due to its dependency on the legislation api (also added currently unused backfill legislation function to verify the test works on my machine)
- Removed remaining usage of unit.TestCase
- Updated ci so that it builds the remaining lambdas on PRs to ensure they are not broken when it comes to release.

- [ ] Update CHANGELOG.md


src/tests/lambda_tests/enrichment_lambda/test_regression_xml_equality.py
```diff
@@ -0,0 +1,108 @@
+# from pathlib import Path
```
anthonyhashemi, line 1: want to discuss this test and the whole legislation API

src/lambdas/enrichment_lambda/steps.py
```diff
@@ -0,0 +1,181 @@
...
+
+    pre_header, end_header_tag, post_header = split_text_by_closing_header_tag(cleaned_file_content)
+
+    replaced_post_header_content = apply_replacements(post_header, replacement_patterns)
+    LOGGER.info("Got post-header replacement text content")
+
+    full_replaced_text_content = pre_header + end_header_tag + replaced_post_header_content
+
+    # raises an lxml.etree.XMLSyntaxError if the output is not valid XML
+    lxml.etree.fromstring(full_replaced_text_content.encode("utf-8"))
```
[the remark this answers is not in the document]
anthonyhashemi: from before but can make a followup commit once first review is done

.github/workflows/deploy_single_environment.yml
```diff
@@ -22,10 +22,10 @@ jobs:
...
         with:
-          ref: ${{ github.event.pull_request.head.sha }}
+          ref: ${{ github.sha }}
           persist-credentials: false
 
-      - name: Set up Python 3.12
+      - name: Set up Python 3.13
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version: "3.12"
```
dragon-dxw: This isn't consistent with the name above.

dragon-dxw (approved):

src/lambdas/enrichment_lambda/index.py
```diff
@@ -0,0 +1,96 @@
...
+from utils.custom_types import APIEndpointBaseURL, DocumentAsXMLString
+from utils.environment_helpers import validate_env_variable
+from utils.helper import parse_file
+
+LOGGER = logging.getLogger()
+LOGGER.setLevel(logging.INFO)
+
+
+def enrich_xml_file(xml: str, enrichment_version: str = "7.4.0") -> str:
+    rules_bucket = validate_env_variable("RULES_FILE_BUCKET")
```
anthonyhashemi: Would be good to add a arg, which when set writes each step's output  to a file. 

Should add some minimal logging in between each step for progress / sanity checks

src/lambdas/enrichment_lambda/steps.py
```diff
@@ -0,0 +1,181 @@
...
+    soup.FRBRManifestation.FRBRdate.insert_after(enriched_date)
+
+    return DocumentAsXMLString(str(soup))
+
+
+def _init_caselaw_nlp(rules_bucket: str, rules_key: str):
+    nlp = spacy.load("en_core_web_sm", exclude=["tok2vec", "attribute_ruler", "lemmatizer", "ner"])
+    nlp.max_length = 5000000
+
+    s3 = boto3.client("s3")
```
anthonyhashemi: we should pull this out and use dependncy injection rather than calling external api deeply nested inside the stack

anthonyhashemi: done

src/lambdas/enrichment_lambda/index.py
anthonyhashemi: added logs for now

.github/workflows/cd_deploy_staging_then_prod.yml
```diff
@@ -27,16 +27,16 @@ jobs:
...
-    with:
-      environment: production
-    secrets:
-      aws_oidc_role_arn: ${{ secrets.AWS_OIDC_ROLE_ARN_PROD }}
-      account_id: ${{ secrets.PROD_ACCOUNT_ID }}
-      workspace: tna-prod
-      app_env: production
-    permissions:
-      id-token: write
+  # deploy_production:
```
anthonyhashemi, line 30: disabled for now until we verify staging is deploying and working fine with this new architecture again. 
Probably want a e2e test running on staging also before auto push to prod.

src/lambdas/enrichment_lambda/index.py
```diff
@@ -57,19 +55,19 @@ def enrich_xml_file(xml: str, enrichment_version: str = "7.4.0") -> str:
 
 
 def process_event(
```
dragon-dxw: I think 'event' might be misleading at this point. This might be "enrich_document"?

src/tests/lambda_tests/enrichment_lambda/test_index.py
```diff
@@ -51,60 +54,87 @@ def test_make_replacements_input_serializes_all_replacement_lists():
...
     mock_fetch.assert_called_once_with(endpoint, "uksc/2024/1", "user", "pass")
     mock_lock.assert_called_once_with(endpoint, "uksc/2024/1", "user", "pass")
-    mock_enrich.assert_called_once_with("<xml/>", enrichment_version="7.4.0")
+    mock_enrich.assert_called_once_with(
+        "<xml/>",
+        [
+            {"pattern": "value"},
+            {"pattern2": "value2"},
+        ],
+        enrichment_version="7.4.0",
```
dragon-dxw: I don't love these needing to be changed every time. Perhaps we should use a version variable or just typing.ANY

docs/vcite.md
```diff
@@ -12,4 +12,8 @@ The vCite on/off control is managed by AWS Parameter Store, which holds a single
 
 Or recommendation is that temporary changes to vCite's on or off state (for example, when processing large batch jobs) should be handled manually in AWS Parameter Store. More permanent changes in state should be ideally be managed in the Terraform code. The reason for this is that if the code is modified and resources the resources are re-built the vCite parameter will, by default, be set back to `off`.
 
-The process at enrichment time is managed by the [xml-validate](/lambdas/xml_validate/index.py) lambda, which is responsible for the checking the value of the vCite parameter.
+The process at enrichment time is managed by the [enrichment lambda](/src/lambdas/enrichment_lambda/index.py), which checks the value of the `vCite` parameter.
+
+When `vCite` is `off`, the enrichment lambda applies JEP enrichment and patches directly back to the API.
+
+When `vCite` is `on`, the enrichment lambda applies JEP enrichment, writes the XML to the external vCite input bucket, and waits for the callback file written to the JEP vCite enriched bucket. The callback object creation event triggers the enrichment lambda again, which then patches the returned XML back to the API.
```
dragon-dxw, line 21: Might be worth adding a warning that this hasn't been used in years.

pyproject.toml
```diff
@@ -46,6 +46,7 @@ sqlalchemy = "2.0.51"
 # Lambda-specific dependency groups (supplementing top-level dependencies)
 [tool.poetry.group.enrichment-lambda.dependencies]
 spacy = "3.8.14"
+click = "8.1.8"
```
dragon-dxw, line 59: Worth adding a comment that this is because of https://github.com/explosion/spaCy/issues/13971 and is "temporary"

dragon-dxw (approved):

src/lambdas/enrichment_lambda/index.py
anthonyhashemi: good point, this has changed singinificantly from processing an event

src/tests/lambda_tests/enrichment_lambda/test_index.py
anthonyhashemi: Yeah it's not ideal but given that we currently rely on this version being set correctly, I think it is beneficial for now. Can revisit once we have the pipeline up and running again

terraform/modules/lambda_s3/lambda.tf
```diff
@@ -1,1362 +1,68 @@
...
-  name = "${local.name}-ecr-repository-rules-update-${local.environment}"
-
-  image_scanning_configuration {
-    scan_on_push = true
-  }
+# ============================================================
+# ECR REPOSITORIES  (one per deployed lambda)
+# ============================================================
 
+resource "aws_ecr_repository" "enrichment" {
```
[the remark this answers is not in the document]
anthonyhashemi: #wiz_ignore same pattern as before. will fix in followup.

src/lambdas/enrichment_lambda/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM public.ecr.aws/lambda/python:3.13@sha256:6b29a28261b892afb364d5246f1d9c6c87e46afc4c99af91635dc4fad6c227f9
```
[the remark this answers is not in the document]
anthonyhashemi: #wiz_ignore same pattern as before and is a lambda, will ignore properly in followup.

PR #1973, opened 2026-06-15
Standardise test infrastructure using Docker/Testcontainers

anthonyhashemi:
## Changes in this PR:

- Replaced testing.postgresql with testcontainers.postgres
- Standardised DB test setup via shared conftest.py fixtures (db_engine, db_connection, etc.)
- Added Docker requirement in CI and setup Buildx for container support
- Introduced make test as the single entry point for running tests
- Removed legacy test scripts and ad-hoc test runners
- Cleaned up outdated DB/test dependencies

**Why**

This change was driven by repeated issues running testing.postgresql locally on macOS, in particular persistent SHM/shared memory errors that were hard to debug and inconsistent across runs.

More broadly, the old setup relied on legacy tooling (testing.postgresql) that behaved differently across environments, making tests brittle and CI/local results inconsistent.

This replaces that approach with a single, Docker-based model so tests are predictable and behave the same everywhere.

**Impact**
- Fixes macOS SHM-related test failures: https://dba.stackexchange.com/questions/275378/dev-shm-size-recommendation-for-postgres-database-in-docker
- Removes environment-specific test behaviour
- Aligns CI and local development

- [ ] Update CHANGELOG.md


README.md
```diff
@@ -15,12 +15,63 @@ Marks up judgments in [Find Case Law](https://caselaw.nationalarchives.gov.uk) w
...
+### Database-backed tests
+
+Some tests require a PostgreSQL database. This is handled automatically using
+Testcontainers:
+
+Each test (or test fixture) starts a temporary PostgreSQL container
+The container is created and destroyed automatically during the test run
+No local PostgreSQL installation is required
+
+#### Prerequisites
```
dancs-dev, line 39: **suggestion (non-blocking):** On Mac, tests may fail unless libpq is installed and on PATH.

I had to run:

```bash
brew install libpq
PATH="/opt/homebrew/opt/libpq/bin:$PATH" && make test
```

It might be worth adding this to the README.

pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 version="6.0.1"
 name="data-enrichment"
-requires-python = ">=3.12,<3.13"
+requires-python = ">=3.13,<3.14"
```
dancs-dev, line 4: **question (non-blocking):** out of curiosity, is there a particular reason Python 14 isn't compatible?

dancs-dev (approved):
Looks good to me! I've not used Testcontainers before but seems like a neat tool.

README.md
anthonyhashemi: oh cheers for checking - will add!

pyproject.toml
anthonyhashemi: To be honest, I'm not sure yet. We have been having issues with the current builds in main and keeping it restricted to one version for now simplifies things. If we get this PR through and deployment working again, we can consider opening this up.

PR #2066, opened 2026-07-02
Full enrichment e2e test

anthonyhashemi:
## Changes in this PR:

- Fix hash of xml test fixtures
- Full enrichment e2e test which runs after staging deploys

Have run in both local_handler (with ssm port forwarding to access db) and sqs modes from my local machine against staging and tests pass. You see the document get created, locked and then enriched, and then deleted.

### DIFF
--- a/src/tests/end_to_end_tests/test_marklogic_e2e.py
+++ b/src/tests/end_to_end_tests/test_marklogic_e2e.py
@@ -1,0 +1,383 @@
+import json
+import logging
+import os
+import re
+import time
+from dataclasses import dataclass
+from pathlib import Path
+
+import boto3
+import lxml.etree
+import pytest
+from caselawclient.Client import DEFAULT_USER_AGENT, MarklogicApiClient
+from caselawclient.models.documents import DocumentURIString
+from caselawclient.models.documents.versions import VersionAnnotation, VersionType
+from caselawclient.models.judgments import Judgment
+from dotenv import load_dotenv
+
+from lambdas.enrichment_lambda.api import (
+    fetch_judgment,
+    lock_judgment,
+    patch_judgment,
+)
+from lambdas.enrichment_lambda.index import handler
+from utils.custom_types import APIEndpointBaseURL
+
+LOGGER = logging.getLogger(__name__)
+
+
+# Load project root .env first, then E2E-specific overrides if present.
+# Assuming this file's path is: <repo>/src/tests/end_to_end_tests/test_marklogic_e2e.py
+PROJECT_ROOT = Path(__file__).resolve().parents[3]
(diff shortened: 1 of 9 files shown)


src/tests/end_to_end_tests/test_marklogic_e2e.py
```diff
@@ -0,0 +1,383 @@
...
+    patch_judgment,
+)
+from lambdas.enrichment_lambda.index import handler
+from utils.custom_types import APIEndpointBaseURL
+
+LOGGER = logging.getLogger(__name__)
+
+
+# Load project root .env first, then E2E-specific overrides if present.
+# File path is: <repo>/src/tests/end_to_end_tests/test_marklogic_e2e.py
```
dragon-dxw: recommend "This file's path is"

src/tests/end_to_end_tests/test_marklogic_e2e.py
```diff
@@ -0,0 +1,383 @@
...
+    )
+    marklogic_client.insert_document_xml(target, xml_element, Judgment, annotation)
+
+    # Set published property to false so Priv API can fetch unpublished docs
+    marklogic_client.set_property(target, "published", "false")
+
+    return True
+
+
+def _delete_fixture_document_if_created(target_uri: str, fixture_was_created: bool) -> None:
```
dragon-dxw, line 160: Do we run this even if the tests are terminated early?

src/tests/end_to_end_tests/test_marklogic_e2e.py
```diff
@@ -0,0 +1,383 @@
...
+def test_enrichment_lambda_triggers_and_updates_marklogic(monkeypatch):
+    if os.getenv("E2E_RUN") != "true":
+        pytest.skip("Set E2E_RUN=true to run E2E tests")
+
+    cfg = E2EConfig.from_env()
+    _configure_enrichment_handler_env(monkeypatch, cfg)
+
+    fixture_was_created = False
+    if cfg.seed_if_missing:
+        fixture_was_created = _seed_fixture_document_from_repo_xml(cfg.uri)
```
dragon-dxw: I think we should consider moving the try statment back to here since there's an event here that might require cleanup?

dragon-dxw (approved):

src/tests/end_to_end_tests/test_marklogic_e2e.py
anthonyhashemi: no, good point!

src/tests/end_to_end_tests/test_marklogic_e2e.py
anthonyhashemi: done

PR #2067, opened 2026-07-02
fix(ci): disable fixture seeding in staging e2e verify for GitHub-hosted runner and reduce duplication of env var reading in test

anthonyhashemi:
## Changes in this PR:

fix(ci): disable fixture seeding in staging e2e verify for GitHub-hosted runner and reduce duplication of env var reading in test

since currently we have no way to access Marklogic via the github runner to seed a test fixture. So we rely upon an existing file in the db until we figure out a more stable longer term solution

### DIFF
--- a/.github/workflows/cd_deploy_staging_then_prod.yml
+++ b/.github/workflows/cd_deploy_staging_then_prod.yml
@@ -66,13 +66,11 @@ jobs:
 
       - name: Run E2E verification in SQS mode
         env:
-          E2E_RUN: "true"
           E2E_TRIGGER_MODE: sqs
-          E2E_SEED_IF_MISSING: "true"
+          E2E_SEED_IF_MISSING: "false"
           E2E_RESTORE_ORIGINAL: "false"
           E2E_TIMEOUT_SECONDS: "300"
-          E2E_SOURCE_XML_PATH: test_files/ewca_civ_2025_673-original.xml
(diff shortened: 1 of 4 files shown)


.github/workflows/cd_deploy_staging_then_prod.yml
```diff
@@ -81,8 +79,6 @@ jobs:
           RULES_FILE_KEY: citation_patterns.jsonl
           VCITE_BUCKET: vcite-tna-files
           VCITE_ENRICHED_BUCKET: foo
-          MARKLOGIC_API_CLIENT_HOST: ${{ secrets.E2E_MARKLOGIC_API_CLIENT_HOST }}
-          MARKLOGIC_USE_HTTPS: "false"
```
dragon-dxw, line 85: Given we'll want to restore most of these, I'm pro-commenting out.

dragon-dxw (approved):

.github/workflows/cd_deploy_staging_then_prod.yml
anthonyhashemi: I don't think it's worth keeping in here, especially since we may not get MarkLogic connectivity it sorted straight away, and it just bloats the workflow. Once MarkLogic connectivity is sorted, reintroducing these is trivial.

